PR #2, opened 2019-10-01
Data models terminology

lnielsen:


rdm/0002-data-models-terminology.md
```diff
@@ -0,0 +1,38 @@
...
+
+You may often see the words like records, record type, schema and data model used interchangeably in the context of Invenio, therefore it’s useful to distinguish what is meant with each term:
+
+- **Record:** a JSON document describing an object.
+- **Record type:** a category of JSON documents describing a specific type of objects (examples: grants, licenses, authors, research outputs).
+- **Record schema:** a JSONSchema describing the JSON structure of a record belonging to a specific record type. A record type may have one or more record schemas (examples: authors schema v1.0, authors schema v1.1, ...).
+- **Deposit record:** a temporary record (e.g. copy of an existing record) used for supporting long running edit sessions that allow users to save their partial changes to a record. A deposit record belongs to the same record type as the source record.
```
fenekku: I am used to it now, but I remember the difficulty I had in grasping this concept because of `invenio-deposit`'s implementation. In `invenio-deposit`, a `Deposit` has both the "draft" feature outlined here **AND** the file association feature. Was something a "deposit" because it had files (files are deposited) or because it was a draft? There might be some disambiguation needed here.

rdm/0002-data-models-terminology.md
```diff
@@ -0,0 +1,38 @@
...
+- **Data model:** a full implementation in Invenio of a specific record type which includes e.g.:
+    - Record schemas describing the internal JSON structure of the records and deposit records.
+    - Search mappings describing how to index records in the search engine.
```
fenekku: We probably want `Record Mapping` as a defined term, right?

rdm/0002-data-models-terminology.md
```diff
@@ -0,0 +1,38 @@
...
+    - Persistent identifier management for records.
+    - Exposure of the records in the REST API and on landing pages.
+    - Deposit support and access control.
+- **Record type category:** a division of record types having shared characteristics (examples: bibliographic record types, authority record types).
+- **Data model category:** a division of data models having shared characteristics (examples: bibliographic data models, authority data models).
+- **Metadata standard/schema:** a standardized schema for metadata (examples: DataCite, Dublin Core, MARC21).
+
+## Drawbacks
+
+Librarians are not using the term **record** as defined as above. Normally a librarian would use the term record to mean a bibliographic record. The term record might better be named a document, however the term is already used so many places through out the Invenio codebase that it will be hard to change.
```
fenekku:
```suggestion
Librarians are not using the term "record" as defined as above. Normally a librarian would use "record" to mean a *bibliographic* record. The term "record" might better be named "document", however the term is already used in so many places through out the Invenio codebase that it will be hard to change.
```

rdm/0002-data-models-terminology.md
```diff
@@ -0,0 +1,38 @@
...
+
+## Motivation
+
+The purpose of this RFC is to disambiguate terms like records, record type, schema and data model that are often used interchangeably in the context of Invenio.
+
+## Detailed design
+
+The central part of InvenioRDM is the primary data model for storing a bibliographic record type describing the digital resources (datasets, software, articles, ...). In addition to the primary data model, InvenioRDM can be extended with one or more secondary data models for storing authority record types (grants, licenses, ...).
+
+You may often see the words like records, record type, schema and data model used interchangeably in the context of Invenio, therefore it’s useful to distinguish what is meant with each term:
```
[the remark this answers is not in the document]
jbenito3: 👍

rdm/0002-data-models-terminology.md
```diff
@@ -0,0 +1,38 @@
...
+
+Definition of terms used when speaking about records and data models.
+
+## Motivation
+
+The purpose of this RFC is to disambiguate terms like records, record type, schema and data model that are often used interchangeably in the context of Invenio.
+
+## Detailed design
+
+The central part of InvenioRDM is the primary data model for storing a bibliographic record type describing the digital resources (datasets, software, articles, ...). In addition to the primary data model, InvenioRDM can be extended with one or more secondary data models for storing authority record types (grants, licenses, ...).
```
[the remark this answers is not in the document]
jbenito3: @fenekku you are right it is more accurate like that. 
On the other hand, I am not sure if I would add the custom data model fields here; just to keep the already complex explanations simple. The main message is, there 2 data models: 1 primary (bibliographic record) and >=0 secondary data models (authorities).  Then, whether they are extensible or not is a different business :) What do you think?

rdm/0002-data-models-terminology.md
```diff
@@ -0,0 +1,38 @@
...
+
+You may often see the words like records, record type, schema and data model used interchangeably in the context of Invenio, therefore it’s useful to distinguish what is meant with each term:
+
+- **Record:** a JSON document describing an object.
+- **Record type:** a category of JSON documents describing a specific type of objects (examples: grants, licenses, authors, research outputs).
```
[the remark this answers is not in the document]
jbenito3: Actually I think it would help. But we should add some clear examples. I was wondering the same thing...isn't record type 'Software'? but of course, it is not, it is a resource type.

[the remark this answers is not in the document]
jbenito3: Maybe we could add something like "Do not confuse with record's resource type, e.g. paper, poster, software, etc. All of them belong to the same record type 'bibliographic record'...Or something similar.

rdm/0002-data-models-terminology.md
jbenito3: > A deposit record belongs to the same record type as the source record.

Shall we say that there is only one 'record type' for deposits? Authorities do not have deposits.

rdm/0002-data-models-terminology.md
```diff
@@ -0,0 +1,38 @@
...
+- **Record schema:** a JSONSchema describing the JSON structure of a record belonging to a specific record type. A record type may have one or more record schemas (examples: authors schema v1.0, authors schema v1.1, ...).
+- **Deposit record:** a temporary record (e.g. copy of an existing record) used for supporting long running edit sessions that allow users to save their partial changes to a record. A deposit record belongs to the same record type as the source record.
+- **Data model:** a full implementation in Invenio of a specific record type which includes e.g.:
+    - Record schemas describing the internal JSON structure of the records and deposit records.
+    - Search mappings describing how to index records in the search engine.
+    - Serializers/loaders used for transforming the internal record to external representations (such as Dublin Core XML, JSON-LD, ...) and vice-versa.
+    - Persistent identifier management for records.
+    - Exposure of the records in the REST API and on landing pages.
```
jbenito3: I do not understand this point very well. I mean, I understand it but not in the context of data model.

jbenito3: all in all it is a bit strange that `data model` is more a set of features than the traditional description of `data model`.

rdm/0002-data-models-terminology.md
```diff
@@ -0,0 +1,38 @@
...
+    - Deposit support and access control.
+- **Record type category:** a division of record types having shared characteristics (examples: bibliographic record types, authority record types).
```
jbenito3: those examples are the *only* examples, right? Shall we be specific on that?

rdm/0002-data-models-terminology.md
```diff
@@ -0,0 +1,38 @@
...
+- **Data model category:** a division of data models having shared characteristics (examples: bibliographic data models, authority data models).
```
jbenito3: Same comment as for **Record type category**

rdm/0002-data-models-terminology.md
```diff
@@ -0,0 +1,38 @@
...
+
+The purpose of this RFC is to disambiguate terms like records, record type, schema and data model that are often used interchangeably in the context of Invenio.
+
+## Detailed design
+
+The central part of InvenioRDM is the primary data model for storing a bibliographic record type describing the digital resources (datasets, software, articles, ...). In addition to the primary data model, InvenioRDM can be extended with one or more secondary data models for storing authority record types (grants, licenses, ...).
+
+You may often see the words like records, record type, schema and data model used interchangeably in the context of Invenio, therefore it’s useful to distinguish what is meant with each term:
+
+- **Record:** a JSON document describing an object.
```
jbenito3: shall add examples of *object*

rdm/0002-data-models-terminology.md
jbenito3: Would it help (or complicate) to add something like this (or better)?

datamodelterminology

Notice that by doing this...I think I understood that record and data model record type categories are the same, nope @lnielsen ?

jbenito3: of course the "..." must be completed with all the elements.

rdm/0002-data-models-terminology.md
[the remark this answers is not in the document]
fenekku: I do want to mention the "data model extension" because the RFC tries to lay out the terms we will be using and we want a term to refer to that concept of extending/customizing the primary data model. How about putting it in a different paragraph like so:

 ```suggestion
The central part of InvenioRDM is the primary data model for storing a bibliographic record type describing the digital resources (datasets, software, articles, ...). In addition, InvenioRDM supports one or more separate secondary data models for storing authority record types (grants, licenses, ...).

The primary data model can be extended with flavour-specific fields that we refer to as data model extensions.
```

?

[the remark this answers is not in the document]
jbenito3: It looks better! We use to say the word "flavour" to describe different apps built with Invenio, for instance: RDM, ILS, Multimedia Archive. It is not the case now I think so if everybody agrees to use it here as well, no problem with me. Otherwise we could simply say "custom fields", "per-case custom fields" , or something similar.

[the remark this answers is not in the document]
fenekku: Ah, we have the flavour / profile RFC [here](https://github.com/inveniosoftware/rfcs/pull/6) for you kind sir! So "flavour" here has a specific meaning.

lnielsen:
After long conversations yesterday:

- Record Type: further discussion needed if this is the right term.
- Data model: further discussion needed. Data model could be understood as the union of all record types (e.g. authors, grants, records, ...). Suggested other names could include: "Record Module" since it is a python module we are talking about.

fenekku:
After long conversations.... more conversations are needed :sweat_smile: 

- Record Type: Resource/Record type issue I am guessing?
- Data model: I like "Record module" because the module does encompass the jsonschema, the  mapping, the marshmallow and even the `Record` api class.    
- Deposit record -> Draft record :100: Ah, that removes a mental load.

lnielsen:
Closing PR as it's old and there's in the mean time been documentation added to InvenioRDM docs which I think covers the same.

PR #3, opened 2019-10-01
WIP: Persistent identifiers for RDM

lnielsen:


rdm/0003-persistent-identifiers.md
```diff
@@ -0,0 +1,105 @@
...
+#### One internal PID per record
+All records must have an internal persistent identifier that is unique for the given data model. Assigning an internal persistent identifier ensures that all records have an identifier from the same identifier scheme. This is useful e.g. for identifier resolution in the REST API as well as for managing relationships internally.
+
+InvenioRDM should further allow administrators to easily configure which internal identifier scheme should be used. The only two identifier schemes provided by default are:
+
+- Record identifier (randomly generated 8 character string)
+- Recid (Invenio v1 style auto-incrementing integers)
+
+#### Internal PID: Record identifier (random number encoded as 8 characters)
+The internal persistent identifier is named the record identifier. The *record identifier* is a randomly generated number encoded using base32. The result is an 8 character string where the last character is a checksum. Following is an example of a record identifier:
```
tmorrell: DataCite is using a mod 97-10, ISO 7064 checksum as implemented in https://github.com/datacite/base32-url. I haven't found a Python implementation, but it probably makes sense to be consistent.

fenekku:
@lnielsen Since this is marked as WIP, would you rather we wait for review?

rdm/0003-persistent-identifiers.md
```diff
@@ -0,0 +1,105 @@
...
+# Persistent identifiers
+
+## Summary
+
+Overall design for management of persistent identifiers in Invenio RDM.
+
+## Motivation
+
+- As a user, I want to have a DOI for my upload, so that I I can cite my research.
+- As a admin, I want to manage multiple DOI prefixes, so that I I don’t need a repository installation per prefix.
```
fenekku:
```suggestion
- As an admin, I want to manage multiple DOI prefixes, so that I don’t need a repository installation per prefix.
```

(The usage of `a` or `an` is a bit of an irregular mess -see what I did there? See https://english.stackexchange.com/questions/152/when-should-i-use-a-vs-an . Turns out I was probably wrong in thinking that "a RFC" was correct. Seem like "an RFC" works because the "R" there is pronounced like a vowel sound - "Arrrr" like a pirate. The "r" in request is not pronounced that way, however; so would require "an". And "user" is an exception to the rule: "a user" is correct. Go figure!).

More on point, when we say "manage multiple DOI prefixes", we mean accept DOI with different prefixes. But do we also mean, allow generation of DOI with different prefixes? Through a potential drop-down in the form for instance?

rdm/0003-persistent-identifiers.md
```diff
@@ -0,0 +1,105 @@
...
+- As a admin, I want to assign both DOIs and Handles for uploads to my repository, so that I I can fulfill business requirements.
+- As a admin, I want to have a Handle for one upload and a DOI for another upload, so that I I don’t need a repository installation per PID scheme.
+- As a user, I want to share a dataset that already have a DOI, so that I I can collect all my research one place.
```
fenekku:
```suggestion
- As a user, I want to share a dataset that already has a DOI, so that I can collect all my research in one place.
```

rdm/0003-persistent-identifiers.md
```diff
@@ -0,0 +1,105 @@
...
+- As a admin, I want to have tombstone pages, so that I I can remove content.
+- As a admin, I want to take over management of a dataset with a DOI from another repository, so that I records can be migrated between repositories..
```
fenekku:
```suggestion
- As an admin, I want to take over management of a dataset with a DOI from another repository, so that records can be migrated between repositories..
```

rdm/0003-persistent-identifiers.md
```diff
@@ -0,0 +1,105 @@
...
+- As a admin, I want to customize which internal PID scheme is used, so that I I can continue my existing identifier scheme..
+- As a admin, I want to ensure all records have a DOI, so that I all records have an DOI.
```
fenekku: We want records to have a DOI because... they should have a DOI :smile_cat: ?

rdm/0003-persistent-identifiers.md
fenekku: We have now implemented this in inveniosoftware/base32-lib . mod 97-10 is used there and it appends 2 checksum digits at the end of the string. To have a large record identifier pool with checksums, maybe we should elevate the size to 10 characters (with the last 2 being checksums), effectively giving us (2^5)^8 = 2^40 =~ 10^12 =~ a trilion possible records? Example being:

```
3sbk2-5j060
```

rdm/0003-persistent-identifiers.md
```diff
@@ -0,0 +1,105 @@
...
+InvenioRDM will initially come with built-in support for DOIs as external identifiers. Further identifier schemes may be provided later. The architecture however should from the beginning allow multiple PID schemes.
+
+#### Managed vs unmanaged PIDs
+
+The external persistent identifiers can either be *managed* or *unmanaged*.
+
+- **Managed PIDs** means that it is the repository’s job to register, update and maintain the PID in an external registry.
+- **Unmanaged PIDs** means that another repository takes care of registering, updating and maintaining the PID in an external registry.
+
+This allows e.g. dataset with an existing DOI to be shared in the repository, with the repository having take control and manage the DOI.
```
fenekku:
```suggestion
This allows e.g. a dataset with an existing DOI to be uploaded in the repository and the repository to take control and manage the DOI. Another dataset with an existing DOI can also be uploaded in the repository without ceding DOI control. Its original DOI provider would still manage the DOI for that dataset.
```

rdm/0003-persistent-identifiers.md
```diff
@@ -0,0 +1,105 @@
...
+
+InvenioRDM must know for each external PID if it’s a managed or unmanaged DOI, since a single DOI from one prefix may be migrated to another repository (i.e. you may or may not manage all DOIs in a prefix).
+
+#### Assignment of identifiers
+A key issue in the management of persistent identifiers is when to assign which external identifier(s) for a given record.
+
+- Internal PIDs are reserved as soon as a deposit record is created.
+- External PIDs are only registered when the deposit record is published.
+
+A record must be able to explicitly defined which external PIDs should be registered with a default behaviour in case nothing is specified.
```
fenekku:
```suggestion
A record must be able to explicitly define which external PIDs should be registered with a default behaviour in case nothing is specified.
```

rdm/0003-persistent-identifiers.md
```diff
@@ -0,0 +1,105 @@
...
+
+#### Tombstones
+
+All persistent identifiers must support having a tombstone pages.
```
fenekku:
```suggestion
All persistent identifiers must support having tombstone pages.
```

rdm/0003-persistent-identifiers.md
```diff
@@ -0,0 +1,105 @@
...
+
+### References
+
+- [1] Fenner (2016). Cool DOIs. https://doi.org/10.5438/55e5-t5c0
+
+## How we teach this
+
+
```
fenekku: This information should be documented/referenced in the invenio-rdm-records / invenio-pidstore.

Our grant work includes contributing writing to a "Reusable Data Best Practices Guidebook". I will translate some of the points made here to this "Guidebook" so that knowledge around identifiers becomes more widespread too.

rdm/0003-persistent-identifiers.md
```diff
@@ -0,0 +1,105 @@
...
+## Drawbacks
+
+
+## Alternatives
```
fenekku: The only "alternative" I remember discussing is having multiple managed DOIs per record. The reason we don't want to go that way, is that it encourages duplication, contributes to things being out-of-sync, and is more complicated (effort).

rdm/0003-persistent-identifiers.md
tmorrell: We've got a use case for generating DOIs with different prefixes (some groups on campus use their own prefixes). How this could be implemented is flexible (different deposit forms, prefix connected to user accounts, drop-down in form).

fenekku: Thanks for the clarification @tmorrell

PR #6, opened 2019-10-02
Customize InvenioRDM product profile

fenekku:


rdm/0006-profiles-and-flavours.md
```diff
@@ -0,0 +1,104 @@
+- Start Date: 2019-10-02
+- RFC PR: [#6](https://github.com/inveniosoftware/rfcs/pull/6)
+- Authors: Guillaume Viger
+
+# Customize created Invenio instance by choosing profile and optionally flavour
+
+## Summary
+
+When running the instance bootstrapping CLI command (`invenio-cli init`), the
+hoster should be given a choice to create an Integrated Library System (ILS) or
```
ppanero: Shall we say ILS, RDM *among others* or for example. My point being, in the future they might come more flavours.

rdm/0006-profiles-and-flavours.md
```diff
@@ -0,0 +1,104 @@
...
+Resource Data Management (RDM) "flavour" Invenio application. If RDM is chosen,
+there are further "profiles" to be chosen from, for instance "default" and
```
ppanero: We sometimes talk about default, generic, institutional repository. I think this is a terminology that might be defined. Shall it go to another RFC?

rdm/0006-profiles-and-flavours.md
```diff
@@ -0,0 +1,104 @@
...
+"biomed". These choices pre-configure the application to varying levels.
+
+## Motivation
+
+> Why are we doing this? What user stories does it support? What is the expected
```
ppanero: For consistency, in other RFCs we do not include the questions. Shall we do it in all? or remove it from all

rdm/0006-profiles-and-flavours.md
```diff
@@ -0,0 +1,104 @@
...
+same settings as other installations in my domain.
+
+As an InvenioRDM (core-, extension-, plugin-)developer for domain stakeholder,
+I want to be able to satisfy the exclusive needs/wants of my specialized
+stakeholders while still being able to develop and use domain authorities **within**
+the InvenioRDM context. I have a dedicated mechanism and space to insert
+customizations shared by instances for my domain this way.
+
+
+## Terminology
```
ppanero: The terminology is used before. Shall it be defined earlier?

rdm/0006-profiles-and-flavours.md
```diff
@@ -0,0 +1,104 @@
...
+
+**flavour**: A *flavour* represents the main use case of the application
+(e.g., ILS or RDM for now) and it translates to a different packaged application.
+The flavour packaged applications are completely different, which calls
+for a different project repository to define them.
+
+**profiles**: *Profiles* are different domains of interest (e.g., Generic or BioMed)
+for the InvenioRDM *flavour* and translate to different configurations,
+data models, and customizations. Those are not major enough to warrant being
```
ppanero: Major enough is correct. However, I believe this is also more of a hierarchy/abstraction layers. If we think about it all profiles could be flavours, but flavours have a high percentage in common. What do you think? I am not sure how to reword this part.

rdm/0006-profiles-and-flavours.md
```diff
@@ -0,0 +1,104 @@
...
+part of different repositories however. The profiles are defined in the same
+repository side-by-side.
+
+
+## Detailed design
+
+```console
+invenio-cli init --flavour RDM --profile biomed
```
ppanero: I was thinking that maybe the profile (or both) could be prometed to the user. Such as:

``` console
$ invenio-cli init
  - Choose your flavour?
  1) RDM
  2) ILS
  - Input choice: 1
  - You have chosen RDM flavour. Which profile do you want?
  1) BioMed
  2) Generic
```

I think is a bit more user friendly, since you prompt the user with the possibilities. Otherwise they need to do a --help to start.

rdm/0006-profiles-and-flavours.md
```diff
@@ -0,0 +1,104 @@
...
+```
+
+(or something close to this syntax) generates the pre-configured RDM application
+with the specific resource types, data model, UI and other configuration relevant
+to Bio-Medical institutions. Passing `--profile generic` generates one with
+default configurations.
+
+The [invenio-app-rdm](https://github.com/inveniosoftware/invenio-app-rdm) and/or
```
ppanero: app-rdm should be a flavour, same level as below ILS

rdm/0006-profiles-and-flavours.md
```diff
@@ -0,0 +1,104 @@
...
+the [invenio-scripts](https://github.com/inveniosoftware/invenio-scripts)
```
ppanero: This has been (in progress) renamed to invenio-cli

rdm/0006-profiles-and-flavours.md
```diff
@@ -0,0 +1,104 @@
...
+As this is the start of the project, what qualifies as profile-worthy
+and what qualifies as flavour-worthy differences will need to be felt out from
+experience. Luckily, at first, only the RDM profile will be supported.
+
+The implied differences between flavours may be overstated and provide decision
+uncertainty as more flavours are added. Lindsay works in a biology research
+institution. She may be unsure about using a "BioMed" profile versus a generic
+one. Should she create her own profile? It should be communicated that
+further customization is possible to match her setting; the decision is not
+irrevocable. Reconfiguration can be done. It should also be made clear, she
```
ppanero: Are we supporting automatic (e.g. command) reconfigure? This should be added to requirements changed in the invenio-cli if so.

rdm/0006-profiles-and-flavours.md
fenekku: I've added a quick summary of the profiles in the terminology to address this.

rdm/0006-profiles-and-flavours.md
fenekku: Forgot it. Removed!

rdm/0006-profiles-and-flavours.md
fenekku: The summary will inevitably use the terms, but I think they are understandable from context there. 

I like presenting definitions only after the motivation has been laid out: the motivation is presented in the problem-domain (plain english) and the definitions are laid out in the solution-domain (technical specifications). In other words, first we motivate the issue in a relatable fashion, then we use specific terms in detailed design to solve it.

rdm/0006-profiles-and-flavours.md
fenekku: Reworded to get at what you are saying. See if that works better.

rdm/0006-profiles-and-flavours.md
fenekku: Will there be both an interactive and non-interactive `init`? We can modify this to fit what the interface the cli ends up using.

rdm/0006-profiles-and-flavours.md
fenekku: I don't think automatic reconfiguration should be supported at this point. Manual reconfiguration should be allowed, but that's just... configuration ;).

rdm/0006-profiles-and-flavours.md
ppanero: Hmm, I haven't thought of it. Currently the cli is saving some config in a `.invenio` hidden file. However, which scenario do you foresee where filling all fields in the config instead of interactive filling them? Some sort of "automated" deployment (e.g. puppet, ansible)?

I would keep it interactive for now. We can get feedback and if people need it implement it. It wouldn't change much current code, mostly add.

rdm/0006-profiles-and-flavours.md
ppanero: Hmm, I haven't thought of it. Currently the cli is saving some config in a `.invenio` hidden file. However, which scenario do you foresee where filling all fields in the config instead of interactive filling them? Some sort of "automated" deployment (e.g. puppet, ansible)?

I would keep it interactive for now. We can get feedback and if people need it implement it. It wouldn't change much current code, mostly add.

rdm/0006-profiles-and-flavours.md
ppanero: Hmm, I haven't thought of it. Currently the cli is saving some config in a .invenio hidden file. However, which scenario do you foresee where filling all fields in the config instead of interactive filling them? Some sort of "automated" deployment (e.g. puppet, ansible)?

I would keep it interactive for now. We can get feedback and if people need it implement it. It wouldn't change much current code, mostly add.

fenekku: Yup, the ansible deployment was what I had in mind or if someone knew what they were doing and wanted to skip the interactivity to just generate what they need.

I will switch to the interactive interface (something like what invenio-cli has) and we can come back to this later.

ppanero: I agree it might be needed. We can add it, and add an issue to invenio-cli. But to be done in the future?

fenekku: Yes, in the future.

rdm/0006-profiles-and-flavours.md
```diff
@@ -0,0 +1,120 @@
...
+As an InvenioRDM (core-, extension-, plugin-)developer for domain stakeholder,
+I want to be able to satisfy the exclusive needs/wants of my specialized
+stakeholders while still being able to develop and use domain authorities **within**
+the InvenioRDM context. I have a dedicated mechanism and space to insert
+customizations shared by instances for my domain this way.
+
+
+## Terminology
+
+**flavour**: A *flavour* represents the main use case of the application
```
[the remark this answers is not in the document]
fenekku: Yes, changing **flavour** for **product** makes sense and is clearer. We could even then have an option for `invenio` and it would just install the `cookiecutter-invenio-instance` . 

I am fine with shifting the focus on the subsequent **repository profiles**, but I'd still keep the (renamed) description for **product** around in order to give context.

fenekku:
Let's use this RFC for laying out how customization of generated product/profile would work in January.

lnielsen:
As discussed IRL, we might not need Product Profiles, so this one is on hold until we really need it.

PR #7, opened 2019-10-02
rdm: access control

ppanero:


rdm/0007-access-control.md
```diff
@@ -0,0 +1,222 @@
...
+Records linked with files (possible for all records) have the additional following actions.
+- Create files: Ability to create files associated with a record. 
+- Read files: Ability to read files associated with a record.
+- Update files: Ability to update files associated with a record.
+- Delete files: Ability to delete files associated with a record.
+- Administrate files: Ability to perform all possible actions on files associated with a record, including reading previous versions of files.
+
+Deposit records have the additional following actions:
+- Publish: Ability to publish a new record.
+- Edit: Ability to edit an already published record. This differs from the record update action.
```
tmorrell: Do you mean an unpublished record? Editing a published record seems to be the same as Updating a record.

rdm/0007-access-control.md
```diff
@@ -0,0 +1,222 @@
...
+- New version: Ability to create a new version of an existing record.
+
+It is possible to extend this model with more actions for each data model as needed.
+
+**Requiring permissions**
+
+A record must store permissions inside the metadata. This can be as simple as a list of owners, or very complex based on custom rules (e.g. books can only be curated by books curators). This rules are translated into `generators`.
```
fenekku: I think the distinction should be made between system-wide policies and record-specific policies i.e. implicit versus explicit permissions.

```suggestion
A record must store permissions inside the metadata. This can be as implicit as relying on a list of owners, or very explicit and detailed based on encoded rules (e.g. books can only be curated by books curators encoded in the record). These rules are translated into `generators`.

Implicit access rules are typically set at the instance-level, while explicit access rules are typically set on a resource by resource basis.
```

Obviously, it's instance-level policies that enable access rules at the resource level...

rdm/0007-access-control.md
```diff
@@ -0,0 +1,222 @@
...
+    _owners: [1, 2, 3]
+
+RecordOwner:
+    needs:
+        return Document._owners
+    excludes:
+        pass
+```
+
+Since the record owner does not set any specific blacklist (*exclude*), the generator would result in an access granted for users with id 1, 2 or 3.
```
fenekku:
```suggestion
Since the record owner does not set any specific blacklist (*excludes*), the generator would result in an access granted for users with id 1, 2 or 3.
```

rdm/0007-access-control.md
```diff
@@ -0,0 +1,222 @@
...
+
+In addition, generators resturn a set of *query filters*, which are used only when searching. The set of rules returned there are in the form of an Elasticsearch Query object. The `RecordOwner` generator would therefore look like:
```
fenekku:
```suggestion
In addition, generators return a set of *query filters*, which are used only when searching. The set of rules returned there are in the form of an Elasticsearch Query object. The `RecordOwner` generator would therefore look like:
```

rdm/0007-access-control.md
```diff
@@ -0,0 +1,222 @@
...
+```
+
+As a result, it can be seen that the *query filter* inverts the paradigm with respect to the *needs*/*excludes*. Instead of receiving a record and returning its owners, it takes the id of the user performing the action and returns a query object. This is due to the nature of the *list* (search) operation. The check must not be *Can user A access record R, for each one of the records in the system*. This would not be scalable. Therefore, the check must be *which documents can be accessed by the user A*.
+
+Invenio RDM comes with a set of predefined generators:
+- Any user: Any user can perform the action.
+- Any user if public: Any user can perform the action if the record is public.
+- Record owners: Owners of the record can access perform the action.
+- Curators: Curators can perform an action.
+- Super Curators: ??
```
fenekku: What is a super curator? Would probably remove for now.

rdm/0007-access-control.md
```diff
@@ -0,0 +1,222 @@
...
+
+
+**Permission Policies**
+
+A *permission policy* defines the set of generators upon which a certain action can be performed. A record policy would be set as:
```
fenekku:
```suggestion
A *permission policy* defines the actions and who can or cannot perform them as generators. A record policy would be set as:
```

rdm/0007-access-control.md
```diff
@@ -0,0 +1,222 @@
...
+
+The policy would return a set of *needs* and *excludes* for each action. The *needs* and *excludes* of a policy is the union of all those of the specified generators.
+
+For the following user story:
+
+```
+As an admin I want to grant read access to the owners of a document, except if the owner is part of team A.
+```
+
+The following policy applies the `RecordOwner` and `TeamMembership` generators to the record read action. The `RecordOwner` generator, allows the owners of the document (defined right above it, with ids 1, 2 and 3) to access the document. On the other hand, `TeamMembershipt` blacklists users belonging to Team A of accesing the document.
```
fenekku:
```suggestion
The following policy applies the `RecordOwner` and `NotTeamAMember` generators to the record read action. The `RecordOwner` generator, allows the owners of the document (defined right above it, with ids 1, 2 and 3) to access the document. On the other hand, `NotTeamAMember` blacklists users belonging to Team A.
```

rdm/0007-access-control.md
```diff
@@ -0,0 +1,222 @@
...
+
+``` python
+Document:
+    _owners: [1, 2, 3]
+
+RecordPermissionPolicy:
+    can_read: [RecordOwner, TeamMembership]
```
fenekku:
```suggestion
    can_read: [RecordOwner, NotTeamAMember]
```

rdm/0007-access-control.md
```diff
@@ -0,0 +1,222 @@
...
+
+RecordOwner:
+    needs:
+        return Document._owners
+
+TeanMembership:
```
fenekku:
```suggestion
NotTeamAMember:
```

rdm/0007-access-control.md
```diff
@@ -0,0 +1,222 @@
...
+    id: 1
+    team: B
+
+UserTwo:
+    id: 2
+    team: A
+```
+
+*UserOne* would be able to read the document since it has id `1`. However, *UserTwo* would not. Eventhough, it has id `2`, it belongs to `Team A` which has been blacklisted.
+
```
fenekku: Having a concrete example like this was really helpful. I am tempted to add  **Example** section to `template.md`.

rdm/0007-access-control.md
```diff
@@ -0,0 +1,222 @@
...
+It might increase the learning curve for current power users. However, for new adopters it will make the access control handling much easier.
+
+## Alternatives
+
+Keep using the current model where every instance has to implement their own using flask-principal.
+
+## Unresolved questions
+
+- Shall we implement *campus user* system role? It would require some sort of IP treatment.
```
fenekku: This was discussed and identified as a good "first pull request effort" by any partner.

rdm/0007-access-control.md
```diff
@@ -0,0 +1,222 @@
...
+- Update and Delete actions over files, are iherited from the associated record. Or are per file?
```
fenekku: I have also had people ask about permission per file. A separate record could be made in this case with different permissions or all files should be given the same permission level as the most restricted file; i.e., per file seems excessive to me for now.

rdm/0007-access-control.md
```diff
@@ -0,0 +1,222 @@
...
+- Deposit Edit vs Record update?
+- Is *query filters* a good name?
+- Mention Elasticsearch Query object. Which is a implementation detail.
+- Is the example with Team A being blacklisted clear? I was thinking for example of ATLAS vs. CMS not allowing each other to access their research (for mission purposes.)
+- Curator vs Supercurator? Where is the line? Community vs Global curator?
+- Are the commented examples good for here? or should they go to the module/RDM docs. (See below)
```
fenekku: My vote is for putting them in the module/RDM docs. It would shorten the RFC and not make any implications/ties to a running service.

rdm/0007-access-control.md
ppanero: Created an issue for it: https://github.com/inveniosoftware/invenio-records-permissions/issues/6

rdm/0007-access-control.md
```diff
@@ -0,0 +1,222 @@
...
+- As a user, I want to grant access to my restricted records by listing specific users and/or groups, so that I can easily revoke the access if needed.
+- As an administrator, I want to predefine roles, so that I can easily keep QA and production accurately reflect each other??
+- As an administrator, I want to restrict access to certain records based on user features (e.g. registered user, member of a specific community, member of academia), so that I can implement my institutions policies.
+- As a user, I want to access my data even after I have left my university.
+- As an administrator, I want to open my repositories for other institutions and their IDMS (?? Institutional Data Management System).
+
+## Terminology
+
+- **Resources**: The term resource in the system represents anything you want to protect. Resources can both be at the object-level (e.g. access to a record) or of more general nature (e.g access to the administration interface).
+Actions: Resources can be protected by requiring the ability to perform one or more actions.
```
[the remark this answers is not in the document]
ppanero: hmm... I think the terminology is still an action. Since we can have `Action('read')`. However, in implementation this is translated to *needs* and *excludes*.

@fenekku what do you think?

rdm/0007-access-control.md
ppanero: I agree on making the distinction. However, isn't the list of owners **explicit** and the ones encoded according to business rules **implicit**?

rdm/0007-access-control.md
ppanero: Hmm... I still like the `TeamMembership`. It is more generic and shows that it could also be used to explicitly accept `TeamB`, `TeamC`. Setting the name as `NotTeamAMember` would make it quite tied to one team use case and would require 2 more generators for the other cases.

rdm/0007-access-control.md
ppanero: Same as before

rdm/0007-access-control.md
ppanero: Same as before

rdm/0007-access-control.md
ppanero: I agree, however, this is only possible on more "implementation" rfcs. Maybe we can add it as optional?

rdm/0007-access-control.md
ppanero: Agree, will leave the short examples above. More in depth go to general docs.

rdm/0007-access-control.md
fenekku: I think the *list* of owners is explicit, but the *associated* permissions are implicit. If I recall correctly, some institutions only want mediated deposits: in this case, being an owner is explicit on the record, but that doesn't give the owner edit permissions because only creators/curators have edit permissions at this institution.

In any case, the permission policy is the ultimate decider: even if there is an explicit encoding in a record, it is the defined generators that interpret/enforce it.

Perhaps the text was describing the effort involved in / look of encoded permissions in metadata? In which case I agree, it can vary from just listing owners to encoding complex rules.

rdm/0007-access-control.md
fenekku: Got it. Let's just make sure `TeamMembership` is used throughout (small typos):

```suggestion
The following policy applies the `RecordOwner` and `TeamMembership` generators to the record read action. The `RecordOwner` generator, allows the owners of the document (defined right above it, with ids 1, 2 and 3) to access the document. On the other hand, `TeamMembership` blacklists users belonging to Team A.
```

rdm/0007-access-control.md
fenekku:
```suggestion
TeamMembership:
```

rdm/0007-access-control.md
```diff
@@ -0,0 +1,190 @@
...
+RecordOwner:
+    needs:
+        return Document._owners
+    excludes:
+        pass
+    query_filter:
+        return Query(owner=user.id)
+```
+
+As a result, it can be seen that the *query filter* inverts the paradigm with respect to the *needs*/*excludes*. Instead of receiving a record and returning its owners, it takes the id of the user performing the action and returns a query object. This is due to the nature of the *list* (search) operation. The check must not be *Can user A access record R, for each one of the records in the system*. This would not be scalable. Therefore, the check must be *which documents can be accessed by the user A*.
```
ntarocco: nit: I would re-phrase the second part `This is due to...` to explain it a little bit better.

rdm/0007-access-control.md
```diff
@@ -0,0 +1,190 @@
...
+
+Invenio RDM comes with a set of predefined generators:
+- Any user: Any user can perform the action.
+- Any user if public: Any user can perform the action if the record is public.
```
ntarocco: How a record is defined `public`?

rdm/0007-access-control.md
```diff
@@ -0,0 +1,190 @@
...
+UserTwo:
+    id: 2
+    team: A
+```
+
+*UserOne* would be able to read the document since it has id `1`. However, *UserTwo* would not. Eventhough, it has id `2`, it belongs to `Team A` which has been blacklisted.
+
+## How we teach this
+
+This is a hard to teach concept because it involves a breaking change on the paradigm as from what it is today. However, it makes it much more user friendly and involves less work to implement custom access control rules. The best way to teach this would be to edit the corresponding section in the Invenio Tutorials (assuming the model is adopted by the whole Invenio Framework). In addition, create tutorial in the general RDM documentation.
```
ntarocco: We should maybe elaborate a little bit more this section.
The module documentation itself should have an extensive documentation since this is a complex topic, with many examples.
I agree about a specific tutorial with many use cases and the main Invenio doc.

rdm/0007-access-control.md
```diff
@@ -0,0 +1,190 @@
...
+
+Keep using the current model where every instance has to implement their own using flask-principal.
+
+## Unresolved questions
+
+- Update and Delete actions over files, are iherited from the associated record. Or are per file?
+- Deposit Edit vs Record update?
+- Is *query filters* a good name?
+- Mention Elasticsearch Query object. Which is a implementation detail.
+- Is the example with Team A being blacklisted clear? I was thinking for example of ATLAS vs. CMS not allowing each other to access their research (for mission purposes.)
```
ntarocco: I think it is clear, but I would mention that it is an example! `Given the following example use case...`

rdm/0007-access-control.md
```diff
@@ -0,0 +1,190 @@
...
+
+## Alternatives
+
+Keep using the current model where every instance has to implement their own using flask-principal.
+
+## Unresolved questions
+
+- Update and Delete actions over files, are iherited from the associated record. Or are per file?
+- Deposit Edit vs Record update?
+- Is *query filters* a good name?
```
ntarocco: I think that it might be too specific to searching/ES, while I would go for something closer to access/REST. But I have no ideas right now! :)

rdm/0007-access-control.md
ppanero: Yup... typo :)

rdm/0007-access-control.md
```diff
@@ -0,0 +1,222 @@
...
+- As a user, I want to restrict access to files in a public record, so that I can safely store and cite my dateset.
+- As a user, I want to restrict files and/or metadata during an embargo period, so that I can comply with publisher rules.
+- As an administrator, I want to have a flexible and easy way of defining access control policies, so that I do not need to code them.
+- As an administrator, I want to have IP-based access-control, so that I can comply with publisher rules.
+- As a user, I want to be able to obtain a sharable link, so that my colleagues can access the files of my restricted record.
+- As an administrator, I want to easily define roles like *academic stuff* based on my IdP (Identity Provider), so that I can implement bussiness rules.
+- As a user, I want to share a preview of a deposit record prior to publishing it, so that my colleagues can validate it.
+- As a user, I want to grant access to my restricted records by listing specific users and/or groups, so that I can easily revoke the access if needed.
+- As an administrator, I want to predefine roles, so that I can easily keep QA and production accurately reflect each other??
```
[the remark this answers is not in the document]
ppanero: Indeed.

rdm/0007-access-control.md
```diff
@@ -0,0 +1,190 @@
...
+
+## Motivation
+
+- As an administrator, I want to have both public and restricted records, so that not all the records are publicly visible.
+- As a user, I want to restrict access to files in a public record, so that I can safely store and cite my dataset.
+- As a user, I want to restrict files and/or metadata during an embargo period, so that I can comply with publisher rules.
+- As an administrator, I want to have a flexible and easy way of defining access control policies, so that I do not need to code them.
+- As an administrator, I want to have IP-based access-control, so that I can comply with publisher rules.
+- As a user, I want to be able to obtain a shareable link, so that my colleagues can access the files of my restricted record.
+- As an administrator, I want to easily define roles like *academic staff* based on my IdP (Identity Provider), so that I can implement business rules.
```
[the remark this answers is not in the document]
Herrner: @ntarocco you mean for IdP based roles? We have the scenario where users will submit to certain communities based on parameters delivered by the IdP (like deanery) as well as the scenario where users are given access to records based on the identity federation (like eduGAIN) they used for signing in.

rdm/0007-access-control.md
```diff
@@ -0,0 +1,190 @@
...
+- As a user, I want to share a preview of a deposit record prior to publishing it, so that my colleagues can validate it.
```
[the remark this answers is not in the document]
ppanero: the previous is about a restricted record. The commented one is about a deposit (not yet published record).

rdm/0007-access-control.md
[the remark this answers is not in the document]
ppanero: Yes, you are right :+1:

rdm/0007-access-control.md
```diff
@@ -0,0 +1,190 @@
...
+- As a user, I want to grant access to my restricted records by listing specific users and/or groups, so that I can easily revoke the access if needed.
+- As an administrator, I want to predefine roles, so that I can easily keep QA and production accurately reflect each other??
+- As an administrator, I want to restrict access to certain records based on user features (e.g. registered user, member of a specific community, member of academia), so that I can implement my institution's policies.
+- As a user, I want to access my data even after I have left my university.
+- As an administrator, I want to open my repositories for other institutions and their IDMS (?? Institutional Data Management System).
+
+## Terminology
+
+- **Resources**: The term resource in the system represents anything you want to protect. Resources can both be at the object-level (e.g. access to a record) or of more general nature (e.g access to the administration interface).
+- **Actions**: Resources can be protected by requiring the ability to perform one or more actions.
```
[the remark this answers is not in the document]
ppanero: I used action, to be consistent with `invenio-access` which has **actions**. I can change it to adhere to the standard and then specify that they are equivalent terms.

rdm/0007-access-control.md
```diff
@@ -0,0 +1,190 @@
...
+
+# Access Control
+
+## Summary
+
+Overall design for management of access control and permissions over resources in Invenio RDM.
+
+## Motivation
+
+- As an administrator, I want to have both public and restricted records, so that not all the records are publicly visible.
```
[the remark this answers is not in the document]
ppanero: rephrased :+1:

rdm/0007-access-control.md
```diff
@@ -0,0 +1,196 @@
...
+
+The policy would return a set of *needs* and *excludes* for each action. The *needs* and *excludes* of a policy is the union of all those of the specified generators.
+
+For the following user story:
+
+```
+As an admin I want to grant read access to the owners of a document, except if the owner is part of team A.
+```
+
+The following policy applies the `RecordOwner` and `TeamMembership` generators to the record read action. The `RecordOwner` generator, allows the owners of the document (defined right above it, with ids 1, 2 and 3) to access the document. On the other hand, `TeamMembership` blacklists users belonging to Team A.
```
lnielsen: I would rename ``TeamMembership`` to ``ExcludeTeam('A')`` or ``ExceptTeam('A')``. This makes reading it easier:

Who can read? record owners except for team A

rdm/0007-access-control.md
```diff
@@ -0,0 +1,196 @@
...
+
+# Access Control
+
+## Summary
+
+Overall design for management of access control and permissions over resources in Invenio RDM. Due to the need of more fine-grained permission policies, the chosen model is ABAC (Attribute Based Access Control).
+
+### Why is this needed
+
+The current implementation (November 2019) requires the re-implementation of full permission factories in order to customize access control. Moreover, it is not clear across implementations if the factories should return a set of needs or simply over-write the `can` function. In the latter and most common does not leverage the whole power of *Flask-Principal*. Therefore, a change of paradigm is needed in order to make access control customizations simple and consistent.
```
fenekku:
```suggestion
The current implementation (November 2019) requires the re-implementation of full permission factories in order to customize access control. Moreover, it is not clear across implementations if the factories should return a set of needs or simply over-write the `can` function. The latter (and most common case) does not leverage the whole power of *Flask-Principal*. Therefore, a change of paradigm is needed in order to make access control customizations simple and consistent.
```

PR #8, opened 2019-10-04
Define naming conventions

fenekku:


framework/0008-naming-conventions.md
```diff
@@ -0,0 +1,71 @@
+- Start Date: 2019-10-03
+- RFC PR: [8](https://github.com/inveniosoftware/rfcs/pull/8)
+- Authors:
+
+# Naming Conventions
+
+> There are only two hard things in Computer Science: cache invalidation and naming things.
+> - Phil Karlton
```
ppanero: :1st_place_medal:

framework/0008-naming-conventions.md
```diff
@@ -0,0 +1,71 @@
...
+- `invenio-previewer` -> `invenio-previewers`
+  ...
+
+Over-specifying naming conventions is restrictive, under-specifying leads
+to ambiguities and muddled names where it is hard to understand what the project
+is about.
+
+## Unresolved questions
+
+Documentation module: should they be `docs-<module>` or `<module>-docs` ? I
```
ppanero: IMHO, I prefer `docs-<module>`

framework/0008-naming-conventions.md
```diff
@@ -0,0 +1,71 @@
...
+## How we teach this
+
+In contributor documentation, we point to this RFC.
+
+
+## Drawbacks
+
+To respect these conventions, a couple of modules would need to be renamed:
+
+- `invenio-deposit` -> `invenio-deposits`
```
ppanero: This would be hard. Maybe it can be done in the future if a refactor is needed. However, we can apply the conventions from now on.

fenekku: I wonder if there is a way on PyPi to have a "package redirect"...

ppanero: Fair, "eventually" is good :angel:. We can open an issue to investigate the package redirect at some point. However, the main issue is the API, whoever programming on invenio sill have `invenio_desposit` everywhere in the imports. But yeah it can be done eventually :+1:

lnielsen:
From discussions in yesterdays RFCs meeting:

RFC is accepted with minor changes (@slint will implement them and merge). 
- Title: Naming Conventions -> Module naming conventions
- How we teach this: new modules will have to go through an RFC, and thus naming conventions can be checked via the RFC process.

PR #10, opened 2019-10-20
framework: SPA auth

ntarocco:


framework/0000-spa-authentication.md
```diff
@@ -0,0 +1,142 @@
...
+success/failure status.
+
+### CSRF
+
+Any HTTP request that sends a payload and relies on cookies authentication should be protected from CSRF attacks.
+The backend, for cookie authenticated requests only (not token based ones), should require and validate the CSRF token
+in the payload or as HTTP header.
+
+All Invenio REST endpoints should be reviewed to require CSRF token.
+Note: this is a **backward incompatible** change!
```
slint: Not 100% sure about that, but I think we could keep backward compatibility since we're talking about a completely new method for auth where a JWT is sent via cookies along with the random CSRF number (e.g. this can bee detected at the beginning of the request by some `app.before_request`). In that case, the CSRF validation doesn't have to happen on each form, but on a pre-processor, meaning we don't have to go through all of the views/forms and mark them.

framework/0000-spa-authentication.md
```diff
@@ -0,0 +1,142 @@
...
+To achieve so, the CSRF validity should be set with longer expiration and a new token should be generated on each
+POST/PUT request to ensure that a fresh one is always available to the client-side app.
+In case the CSRF token validation fails, the backend should provide a new token in the response.
+
+### SameSite cookie
+
+[SameSite cookie](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies) is a feature that allow servers to
+instrument browser to send cookies only to requests to the same site, and not to send them cross-site.
+This will enable protection against CSRF.
+Unfortunately, the `SameSite` feature is relatively new and it is supported only by recent browsers: it cannot
```
slint: I would also add a link to [CanIUse for the feature](https://caniuse.com/#feat=same-site-cookie-attribute)

framework/0000-spa-authentication.md
```diff
@@ -0,0 +1,142 @@
...
+- Authors: Nicola Tarocco
+
+# Enable Single Page app authentication in Invenio
+
+## Summary
+
+Provide a secure authentication flow for Single Page Applications with CSRF support and ensure protection to
+Cross-Site Scripting (XSS) and Cross-Site Request Forgery (CSRF) attacks.
+
+Reference article: https://medium.com/@jcbaey/authentication-in-spa-reactjs-and-vuejs-the-right-way-e4a9ac5cd9a3
```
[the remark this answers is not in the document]
ntarocco: It is already the link of the updated version :) I will add the second link to the cookies split.

framework/0000-spa-authentication.md
ntarocco: Indeed, it was implemented like this, changing the RFC.

PR #12, opened 2019-10-24
rdm: define record permissions schema and logic

fenekku:
This is an extension to RFC #7 . To be verified by implementation work (https://github.com/inveniosoftware/invenio-records-permissions/issues/16).


rdm/0012-access-control-metadata.md
```diff
@@ -0,0 +1,179 @@
...
+schema.
+
+**Interpretation**
+
+Policies and their Generators will be provided out-of-the-box to enforce the
+permissions listed in the record:
+
+For `access_right == "open"`, the `permissions.can_read` is skipped for reading
+and, `permissions.can_update` and `permissions.can_delete` are consulted
+when updating and deleting respectively.
```
fenekku: A bit too purist in practice...

ppanero:
@ntarocco I had the same issue in Citadel Search (rebrand of CERN Search). At the moment (8 months ago) there was no way of removing a field form the search in ES (v6.2), maybe now there is one.

The way I solved it was to replicate the behaviour of ``_all`` field (which was deprecated in ES, maybe we should also check why). So now in my schemas I have something like:

``` json
_access: explicit permissions here
_ data: all fields with content to be searchable
```

Afterwards by using the ES python DSL, you can specify the field in which the query will be performed (https://github.com/elastic/elasticsearch-dsl-py/blob/master/elasticsearch_dsl/query.py#L12 through params to https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-query-string-query.html see "default_field").

However, I recall issues with booleans and dates, because they were not strings so I had to move them out of this default fields.

From my perspective the tasks would be:

- check is fields can be excluded from search queries
- check why `_all` was deprecated
- implement something like the `_data` metioned above.
  - Evaluate `copy_to` vs having a field containing everything e.g. `_data` (`copy_to` this would have implications when doing queries to specific fields, which leads to the next point)
  - Do we need a query parser?

PR #13, opened 2019-11-27
rdm: outline access levels

fenekku:


rdm/0013-access-levels.md
```diff
@@ -0,0 +1,173 @@
...
+
+# Access Levels
+
+## Summary
+
+RFC [#7](https://github.com/inveniosoftware/rfcs/pull/7) covers the overall
+access control strategy and the policy implementation details. In particular,
+it covers low-level permissions such as `can_create`, `can_read`, `can_update`
+and so on. This RFC (#13) complements #7 by outlining access levels (typically) on
+a record. Access levels are user-friendly grouping rich with meaning that
```
ppanero: Minor: 

> Access levels are user-friendly grouping rich

Without a comma afterwards is a bit confusing to me. Had to read it a few times :sweat_smile:

rdm/0013-access-levels.md
```diff
@@ -0,0 +1,173 @@
...
+capture all the low-level permissions an identity has on a record.
+
+## Terminology
+
+**Access level**: A human-friendly grouping of low-level permissions that is tied to an identity for a record. Example: *Metadata Curator*.
```
ppanero: Minor: Limit line to 80 as the rest?

rdm/0013-access-levels.md
```diff
@@ -0,0 +1,173 @@
...
+Metadata Curator access level on his record. What needs to be clear is what
+each access level means, so that the right one can be picked. The UI should
+provide that information when it comes time to choose.
+
+
+## Drawbacks
+
+Because InvenioRDM decides the mapping of access level to permissions, there
+will always be administrators who have different needs. They will have to
+code their own, but the structure allows it.
```
ppanero: The use of a policy allows customization code wise but not naming wise (e.g. `files_curator`), this would require a change in the schema.

Would it be possible to create this levels as a dynamically loaded abstraction (i.e. roles?) that then maps the provides to needs of the datamodel (`owner`, `files_restricted`, etc.).

Nonetheless, for explicit (i.e. sharing link) we do need something like this.

rdm/0013-access-levels.md
fenekku: Reworded differently. Hope it helps.

rdm/0013-access-levels.md
fenekku: done

rdm/0013-access-levels.md
fenekku: The implementation here: https://github.com/inveniosoftware/invenio-records-permissions/pull/29 might shed some light on the dynamic possibilities. Essentially, as long as the access level (whatever name it has) is slotted for the right actions, those will be used in the search without further schema changes.

That being said having greater control over these access levels could be done via a role / database association system... but I would wait for that since it is heavy. Maybe some configuration could go a long way (even then, I am not a big fan of kicking configuration down to the instance administrator).

ppanero: > but I would wait for that since it is heavy

Agree

> even then, I am not a big fan of kicking configuration down to the instance administrator

Also agree.

This was mostly a comment. Let's see what @slint @lnielsen think about it.

fenekku: I think I see what you mean.

The missing piece I haven't put in there, but was going to be added eventually to address that is the mapping from: human-readable access level (e.g.,  `"Metadata Curator"`) to this internal access level key name (`"metadata_curator"`). Then the two can change independently as long as they are connected with each other. I could definitely see `"Metadata Curator"` instead be `"Edit Record Metadata"`. This way we can keep `metadata_curator` internally without the UI affecting the internals.

PR #15, opened 2020-03-12
dataflow: lay out some notes

fenekku:
I was writing this down following some of the discussions we had and didn't know a good place to put it. Not attached to have it here, but just needed a place for conversation... Feel absolutely free to extract it out somewhere. 

This is a super sketch / work in progress. The goals are to get a good sense of where different dataflow operations fit and how to structure the code with concerns laid out as much as possible, The deposit page and the resource controller work will feed and be fed with this discussion.


fenekku:
Superseded by https://github.com/inveniosoftware/rfcs/issues/23 . Closing.

PR #16, opened 2020-03-13
framework: python 2 end of life

lnielsen:


framework/0016-python2.md
```diff
@@ -0,0 +1,148 @@
...
+    python: "2.7"
+    condition: $DEPLOY = true
+```
+
+To
+
+```yaml
+python:
+   - "3.6"
+   - "3.8"
```
slint: Not 3.7?

framework/0016-python2.md
```diff
@@ -0,0 +1,148 @@
...
+
+deploy:
+  # ...
+    python: "3.6"
+    condition: $DEPLOY = true
+```
+
+Make sure you check that the build matrix has the ``$DEPLOY`` included.
```
slint: Travis has [a `skip_existing` flag](https://docs.travis-ci.com/user/deployment/pypi/#upload-artifacts-only-once), which would help with getting rid of the `$DEPLOY` variable. So we could simplify the `deploy` section to:

```yaml
deploy:
  provider: pypi
  user: inveniosoftware
  password:
    secure: "..."
  distributions: "sdist bdist_wheel"
  on:
    tags: true
  skip_existing: true
```

lnielsen: Just checking: We won't need the python tag anymore either as well?

slint: Nope, since it doesn't matter which version makes it first, plus we don't build any version-specific wheels (because we have `universal = 1` in the `setup.cfg`'s `[bdist_wheel]`)

Issue #1, opened 2019-10-01
Decide on decision process for RFCs


lnielsen:
There should be a defined decision process for RFCs to be either accepted or rejected. It's important that the process be **as lightweight as possible** while still allowing some structure. 

The purpose of the RFC is to make it easier to collaborate together and ensure we have some minimal documentation about design decisions taken.

My current thoughts:

- Possibly Invenio Architects could take the final decision based on some principles of e.g. consensus/impact and inline with the code of conduct.
- Decisions of architects could be appealed to the Invenio Governance (which will be defined as part of InvenioRDM).
- There must be a defined timeline for RFCs (e.g. you are guaranteed to get a reply within X amount of time from an architect)
- The process has to somehow also take into consideration the implementation of the RFC.

There should also be a way for how less technical users can get help to write up an RFC.

Please write your ideas/comments/suggestions for how the RFC process should work. Links to other RFC processes are also very welcome.

fenekku:
>  Invenio Architects could take the final decision based on some principles

Yes, we should make sure a person (architect) with a wider perspective / vision on Invenio is in the loop.
 
>  inline with the code of conduct

Was "inline with roadmap" meant here? Phrasing the RFC in a decent manner is of course the encouraged way, but English may not necessarily be the submitter's first language and may explain the tone. Reviewers should account for that (massage or encourage submitter to mollify their language).

> Decisions of architects could be appealed to the Invenio Governance

If the RFC is rejected, then a closing comment with justification has to be provided. Submitters and reviewers probably took time + effort to discuss the problem, so a nice explanation would be appreciated. Also, the RFC or issue may be revisited later and it's important to be able to go back and re-assess the reasoning behind the rejection.

> appealed to the Invenio Governance

Sounds like a court decision :smile: which may be too heavy. *If* the context changes since the initial rejection, then a closed RFC should be fair game to revisit. Should a new RFC be made or old one re-opened and edited? A new commit would help preserve decision history in latter case.       

> defined timeline for RFCs

[...]

PR #17, opened 2020-03-16
[WIP] rdm: Invenio modules for creation and handling of standardized user requests

lnielsen:


rdm/0000-rdm-requests.md
```diff
@@ -0,0 +1,261 @@
+- Start Date: 2020-03-16
+- RFC PR: [#17](https://github.com/inveniosoftware/rfcs/pull/17)
+- Authors: Lars Holm Nielsen
+
```
ppanero: State? To comply with what was said in https://github.com/inveniosoftware/rfcs/issues/1

lnielsen:
It's not supposed to handle normal support tickets. On the other hand, a normal ticket system cannot automate the task handling (i.e go an increase the quota of a bucket). Also, a normal ticket system is minded on a single support entity. Essentially, this can be used to handle a user transferring a record to another user, and doesn't involve the support team.

Also, probably a support system like UserVoice and the like, could probably be integrated in some way with this. I.e. the request opening a ticket in the external system.

PR #35, opened 2020-07-16
framework: error handling and response for REST

lnielsen:
During the implementation of Flask-Resources it became clear that we need to take some choices regarding HTTP error responses from the REST API and how it's implemented.

Error response bodies are not subject to content negotiation and  thus always JSON formatted. Further a simple method for error handling and returning JSON formatted responses is presented. The method relies on Flask error handlers and extension of the HTTPException.

### DIFF
--- a/rfcs/framework-0035-error-handling-and-response.md
+++ b/rfcs/framework-0035-error-handling-and-response.md
@@ -1,0 +1,323 @@
+- Start Date: 2020-07-15
+- RFC PR: [#35](https://github.com/inveniosoftware/rfcs/pull/35)
+- Authors: Lars Holm Nielsen, Zach Zacharodimos
+- State: DRAFT
+
+# REST API error responses in Invenio
+
+## Summary
+
+Error response bodies are not subject to content negotiation and  thus always JSON formatted. Further a simple method for error handling and returning JSON formatted responses is presented. The method relies on Flask error handlers and extension of the HTTPException.
+
+## Motivation
+
+Clear and consistent error handling throughout Invenio REST APIs is highly important for the clients accessing our REST APIs, as well as for properly logging errors and tracking errors. It's further important for a clean code architecture.
+
+
+## Detailed design
+
+The Invenio REST APIs makes use of HTTP Content Negotiation in order to render a record in multiple different formats.
+
+**Example of content negotiation with succcessful response**
+*Request*
+
+```http
+GET /records/1 HTTP/1.1
+Accept: application/xml
+```
+
+*Response*
+
+```http
+HTTP/1.1 200 OK
+Content-Type: application/xml
+...
+
+<record>
+    ...
+</record>
+```
+
+Notice, the Accept and Content-Type header. If you ask for ``application/json`` instead, you could get a JSON-formatted HTTP response body instead of XML-formatted.
+
+
+### Should error responses use content negotiation?
+
+A key question is if an error response should also be subject to HTTP Content Negotiation? For instance, if the following HTTP request results in a HTTP 404 Not Found error, what does the response looks like:
+
+*Request*
+
+```http
+GET /records/1 HTTP/1.1
+Accept: application/xml
+```
+
+*Response example 1 (with content negotiation)*
+
+The client asked for XML so an XML response is returned.
+
+```http
+HTTP/1.1 404 Not Found
+Content-Type: application/xml
+...
+
+<error>
+    <code>400</code>
+    <message>The requested resource was not found.</message>
+</error>
+```
+
+*Response example 2 (without content negotiation)*
+
+The client asked for XML, but instead a JSON response is returned for all errors.
+
+```http
+HTTP/1.1 404 Not Found
+Content-Type: application/json
+...
+
+{
+    "code": 404,
+    "message": "The requested resource was not found."
+}
+```
+
+
+**Error response with content negotiation (example 1)**
+
+At a theoretical level, the above example may seem logical that a client requests XML, and gets an error response in XML. However for other formats, this becomes less clear. For instance, some formats may not easily encode errors - e.g. how would an error response look like for CSV or BibTeX?
+
+Similar, at an architecture-level, an application deals with a wide range of errors, and decoupling the error from the serialization of the response body, means that we need very clear definition of errors so that the error serializers have a limited set of errors as input to deal with, otherwise it becomes and overwhelming task to write an error serializer. Currently, this does not seem feasible in Invenio code base.
+
+Also, at an architecture-level, errors may come from many sources, . Take e.g. a 404 Not Found error. This one can come from the service layer or from the Flask routing layer. This would mean error handling would need to be global for the application in order to have consistent serialization of both errors.
+
+At a client-level, the most important is having a machine-readable format for the error, and especially it's important for Single-Page applications in order to display errors. The response will always have a content type defined, and thus it's reasonable striaghtforward to handle errors in a different format that what was requested, especially if it is only for errors.
+
+**Error response without content negotiation (example 2)**
+
+Having only a single response format for errors simplifies all of above mentioned issues with the content negotiation. Serializers needs only to deal with how to serialize a successful response and not a pletora of errors, and thus becomes easier to write. Having only one format, means errors can easily be generated from multiple different sources and produce the same error (e.g. 404 from service vs routing layer), and clients can easily decode errors and know it's always the same format.
+
+Because of above reasons, it means that *error response bodies are not subject to content negotiation and thus always JSON formatted*
+
+### Who is responsible for rendering the HTTP response body for an error?
+
+Having established that error response are not subject to content negotiation, the next question is who is responsible for rendering the HTTP response body for an error.
+
+As a prerequisite, please see [error handling in Flask](https://flask.palletsprojects.com/en/1.1.x/errorhandling/#error-handlers).
+
+**Sources of errors**
+
+The following are usual sources of errors (i.e. exceptions being raised):
+
+1. Flask layer, for instance routing errors.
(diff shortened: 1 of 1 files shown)


rfcs/framework-0035-error-handling-and-response.md
```diff
@@ -0,0 +1,323 @@
...
+    # A handler was found, use it to return response
+    return handler(e)
+
+def handle_http_exception(e):
+    handler = find_error_handler(e)
+    if handler is None:
+        # No handler was found, but it is an HTTPException.
+        return e # <----- return
+    # A handler was found, use it to return response
+    return handler(e)
```
fenekku, line 171: Not sure I follow why HttpExceptions are treated differently. The pseudo-code above simplifies to:

 ```suggestion
def handle_exception(e):
    handler = find_error_handler(e)
    if handler is None:
        # No handler was found
        raise e  # <----- reraise. This results in a 500 if e is not an HTTPException or 
                 # the appropriate status code because it is an HTTPException
    # A handler was found, use it to return response
    raise handler(e)

```

File: rfcs/framework-0035-error-handling-and-response.md [hunk 2]
```diff
@@ -0,0 +1,323 @@
...
+```
+
+The key in understanding the difference between an normal exception and an HTTPException is that an HTTPException is considered a valid response by Flask's ``make_reponse``. This allows it to be returned by the ``handle_http_exception()`` function.
+
+**Recap**
+
+Source of errors (exceptions raised) is three layers (Flask, View, Service). For handling these errors (i.e. making an HTTP error response) we can use Flask and Werkzeug's error handling system. The view layer is not well-suited for this.
+
```
fenekku, line 179: ∑(O_O;) I think I am not quite convinced that the view layer is not well-suited... Actually, what is puzzling me is that we seem to try to build up the argument for why to use Flask's `error_handler` functionality, but handling errors via a `try-catch` in the decorators of flask-resources or via `error_handler` of Flask (functionality in flask-resources as well) is actually equivalent (or at least the difference hasn't been made clear to me... :sweat_smile: ). In both cases we are placed at the blueprint level...

rfcs/framework-0035-error-handling-and-response.md
```diff
@@ -0,0 +1,323 @@
...
+**Flask application**
+
+Next, we install an error handler on the Flask application for ``HTTPExceptions``:
+
+```python
+from Flask import Flask
+
+app = Flask(__name__)
+
+@app.errorhandler(HTTPException)
```
fenekku, line 213: I think that's the core of my disagreement: in invenio-records-resources this can be done at the app level but in flask-resources this should be done on a blueprint **only**.

rfcs/framework-0035-error-handling-and-response.md
lnielsen: It's because the "pseudo-code" is taken directly from Flask were there are two methods, so I prefer to keep it as close as possible to Flask code.

rfcs/framework-0035-error-handling-and-response.md
```diff
@@ -0,0 +1,323 @@
...
+
+At a client-level, the most important is having a machine-readable format for the error, and especially it's important for Single-Page applications in order to display errors. The response will always have a content type defined, and thus it's reasonable striaghtforward to handle errors in a different format that what was requested, especially if it is only for errors.
+
+**Error response without content negotiation (example 2)**
+
+Having only a single response format for errors simplifies all of above mentioned issues with the content negotiation. Serializers needs only to deal with how to serialize a successful response and not a pletora of errors, and thus becomes easier to write. Having only one format, means errors can easily be generated from multiple different sources and produce the same error (e.g. 404 from service vs routing layer), and clients can easily decode errors and know it's always the same format.
+
+Because of above reasons, it means that *error response bodies are not subject to content negotiation and thus always JSON formatted*
```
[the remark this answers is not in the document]
lnielsen: You have to separate the two cases 1) developing Invenio Framework and 2) using Invenio Framework to build applications.

This RFC only deals with (1). It's not supposed to be prescriptive about (2). If a person using Invenio wants to make a ContentNegotiatingException that's able to reply in XML and JSON it's fine.

I'll try to update the intro to make this clear.

rfcs/framework-0035-error-handling-and-response.md
```diff
@@ -0,0 +1,323 @@
...
+The view layer error handling is best suited for very custom error handling from the service layer. E.g implementation of a custom API with custom reposnses.
+
+The issue with view layer error handling is that only errors from the view and service/data source layer is being caught. Any error in the Flask layer (routing, request processing functions etc) will escape the try/except-clause (because they happen outside the function).
+
+Example of a 404 error:
+
+1. A routing error will result in a 404 error from the Flask layer.
+2. An object that doesn't exists in the database, will likely result in a 404 error from the service layer.
+
+A view layer can only catch the latter. Thus, if we render an error response in the view, we still have to deal with error response for Flask layer errors.
```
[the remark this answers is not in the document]
lnielsen: Where would the 400 error come from in the Flask-layer? Query parsing is a view layer error?

rfcs/framework-0035-error-handling-and-response.md
lnielsen: It's not equivalent because Flask-level errors escape the view layer.

rfcs/framework-0035-error-handling-and-response.md
lnielsen: The RFC is about error handling principles in an Invenio-based application, and it is not about if Flask-Resources or Invenio-Records-Resources should install application-level error handlers. The main purpose is to document the thoughts behind the *current* approach.

Already today, Invenio-REST is installing the application-level error handlers, and Invenio-Records-REST is installing blueprint level error handlers more or less according to what is described here. There are few details in the current implementation that could be simplified and improved, but overall the approach is the documented one. So you can say this RFC is simply confirming the current approach. 

I think it's probably more a misunderstanding of scope that needs to better documented in the RFC.

rfcs/framework-0035-error-handling-and-response.md
```diff
@@ -0,0 +1,323 @@
...
+
+**Sources of errors**
+
+The following are usual sources of errors (i.e. exceptions being raised):
+
+1. Flask layer, for instance routing errors.
+1. View layer, for instance ``abort(400)`` or ``raise BadRequest()``.
+1. Service/data source layer, for instance Elasticsearch ``RequestError`` or ``PIDDeleteError``.
+
+A bit oversimplified, you can say that Flask/view layer errors usaully  result in HTTP errors such as 40x errors (Bad Request, Not Found).
```
ntarocco, line 121:
```suggestion
A bit oversimplified, you can say that Flask/view layer errors usually result in HTTP errors such as 40x errors (Bad Request, Not Found).
```

rfcs/framework-0035-error-handling-and-response.md
lnielsen: What this means is basically you'll need both Flask errors handlers and you view error. Instead, you can simply let the Flask error handler do the entire job and not duplicate error handling.

lnielsen:
I've added a section about scope. Please check and see if this clarifies.

rfcs/framework-0035-error-handling-and-response.md
```diff
@@ -0,0 +1,328 @@
...
+
+# Install handlers on blueprints.
+def register_error_map(blueprint, error_map):
+    for exc_or_code, handler in error_map.items():
+        blueprint.register_error_handler(exc_or_code, handler)
+```
+
+#### Usage
+
+The above error handler technique, can the be used with the existing exceptions and Flask idoms:
```
ntarocco, line 251:
```suggestion
The above error handler technique, can the be used with the existing exceptions and Flask idioms:
```

rfcs/framework-0035-error-handling-and-response.md
```diff
@@ -0,0 +1,328 @@
+- Start Date: 2020-07-15
+- RFC PR: [#35](https://github.com/inveniosoftware/rfcs/pull/35)
+- Authors: Lars Holm Nielsen, Zach Zacharodimos
```
zzacharo, line 3: If you want my official name is `Zacharias Zacharodimos` hehe :)

rfcs/framework-0035-error-handling-and-response.md
```diff
@@ -0,0 +1,328 @@
...
+
+IMPORTANT: As a prerequisite, please read [error handling in Flask](https://flask.palletsprojects.com/en/1.1.x/errorhandling/#error-handlers).
+
+**Sources of errors**
+
+The following are usual sources of errors (i.e. exceptions being raised):
+
+1. Flask layer, for instance routing errors.
+1. View layer, for instance ``abort(400)`` or ``raise BadRequest()``.
+1. Service/data source layer, for instance Elasticsearch ``RequestError`` or ``PIDDeleteError``.
```
zzacharo, line 119: Is it only me or here there 3 `1)`? @lnielsen

rfcs/framework-0035-error-handling-and-response.md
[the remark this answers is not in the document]
fenekku: Right. I think this is sort of dealt with by the possibility of creating custom exceptions that know how to render themselves potentially.

rfcs/framework-0035-error-handling-and-response.md
fenekku: Probably the scope is throwing me off here :sweat_smile: . Is this how we want error handling to be managed in the future (in flask-resources and invenio-records-resources) or is it how it *is* working in production invenio (invenio-rest)? Maybe the RFC cuts things differently and that is not even a concern... :thinking: ?

PR #40, opened 2020-11-23
framework: record relations

lnielsen:
Record relations is conceptually similar to foreign key relationships in relational databases. Record relations are used to model relationships between different entities such as for instance referencing a grant record from a bibliographic record.

In terms of Invenio, the use cases are many. For instance, a bibliographic record could be related to:

- Resource type
- Subjects/keywords
- Languages
- Authors
- Organisations
- Geonames
- ...

[Codimd link](https://codimd.web.cern.ch/XfEvnoWrRKuKhf0N3glfjA?both)

### DIFF
--- a/rfcs/framework-0040-relations.md
+++ b/rfcs/framework-0040-relations.md
@@ -43,7 +43,7 @@ The issues with this approach are:
 
 - Technically diffcult to understand and debug because of the resolving magic happening behind the scenes. A special library, catches requests to special URLs, and translates them into database queries. This is non-intuitive and very difficult to trace when it's not working as expected.
 - Lack of fine grained control. It is not easily possible to control which fields from the referenced record is included. For instance, a biblographic record relates to a grant record that relates to a funder record. All three records will be dereferenced, causing the index JSON to become very big.
-- Security/performance - because JSONRef is based on making HTTP requests, there's a risk that badly formed URLs, will cause HTTP requests to be made. At best this will only cause performance issues, at worst it opens a backdoor for injecting content in the JSON document. While this could also be considered a feature, ability to relate to externally hosted records, we think the performance/security issues makes this technically impossible.
+- Security/performance - because JSONRef is based on making HTTP requests, there's a risk that badly formed URLs, will cause HTTP requests to be made. At best this will only cause performance issues, at worst it opens a backdoor for injecting content in the JSON document. While this could also be considered a feature, ability to relate to externally hosted records, we think the performance/security issues makes this undesirable.
 
 
 ## Design
@@ -70,7 +70,7 @@ We propose the following syntax:
 }
 ```
 
-A relation is an object with a single key ``id`` with a string value for a persistent identifier for the related record.
+A relation is an object with an ``id`` key whose value is the persistent identifier of the related record as a string.
 
 A relation is considered dereferenced if it contains more than one key.
 
@@ -78,7 +78,7 @@ The syntax was chosen for a number of reasons:
 
 - We would like to avoid confusion with JSONRef as it is already implemented in Invenio, and will work differently than JSONRef.
 - We would like a syntax that can be used both in the internal JSON document stored in the database, as well as in the external JSON used in the REST API.
-- By using an object we can easily dereference it. I.e. ``{"id": "..."}`` is dereferenced to ``{"id": "...", "key1": "val1", ...}``. The key ``id`` always exists (unlike e.g. when using JSONRef).
+- By using an object we can easily dereference it i.e. ``{"id": "..."}`` is dereferenced to ``{"id": "...", "key1": "val1", ...}``. The key ``id`` always exists (unlike e.g. when using JSONRef).
 - Relations should be to a persistent identifier for the related record to make the relation as stable as possible. Also, we wanted to avoid having mulit-key objects as possible persistent identifiers to ensure APIs could be as standardized as much as possible.
 - We did not want to include e.g. type information for the related record, to keep the JSON small as well as keep the type information on the field definition itself, instead of inside the referenced object.
 
@@ -143,7 +143,7 @@ Due to the cardinality and concurrency issues, the database should always only s
 
 All records have a version counter for optimistic concurrency control. Thus, if everytime we dereference a relation we also store the version counter of the referenced record, we will be able to reliably detect which records must be dereferenced again.
 
-As searches a likely to happen in Elasticsearch, we suggest to make a single compound key
+As searches are likely to happen in Elasticsearch, we suggest to make a single compound key.
 
 We suggest adding a ``@v`` key in the *dereferenced data* (not in the database) that includes the version counter of the referenced record:
 
@@ -151,7 +151,7 @@ We suggest adding a ``@v`` key in the *dereferenced data* (not in the database) 
 {"id": "1234", "@v": 18, "a": ""}
 ```
 
-It should be investigated if this if sufficient in order to proper queries in Elasticsearch. Alternatively, we will need to create a compound key concatenating the id and the version - for instance:
+It should be investigated if this if sufficient in order to generate proper queries in Elasticsearch. Alternatively, we will need to create a compound key concatenating the id and the version - for instance:
 
 ```json
 {"id": "1234", "@v": "v18:1234", "a": ""}
@@ -167,7 +167,7 @@ Note, we cannot replace the ``id`` key, as it must be stable between the normali
 
 Relations could be defined by a new system field named ``RelationsField``. The reason for using the name "relations" is to avoid confusion with REST API HATEOS links in the responses.
 
-```pythton
+```python
 class Record(RecordBase):
     relations = RelationsField(
         resource_type=PIDRelation('metadata.resource_type', ResourceTypeRecord.pid),
@@ -195,7 +195,7 @@ However, this will make it a bit more diffcult to operate on all relations at on
 
 ##### Limiting attributes
 
-Limting the attributes could be done as:
+Limiting the attributes could be done as:
 
 ```python
 PIDRelation('metadata.resource_type', ResourceTypeRecord.pid, attrs=[
--- a/rfcs/framework-0040-relations.md
+++ b/rfcs/framework-0040-relations.md
@@ -1,0 +1,321 @@
+- Start Date: 2020-11-16
+- RFC PR: [#40](https://github.com/inveniosoftware/rfcs/pull/40)
+- Authors: Lars Holm Nielsen, Nicola Tarocco, Alex Ioannidis
+- State: DRAFT
+
+# Record relations
+
+## Summary
+
+TODO
+
+## Motivation
+
+Record relations is conceptually similar to foreign key relationships in relational databases. Record relations are used to model relationships between different entities such as for instance referencing a grant record from a bibliographic record.
+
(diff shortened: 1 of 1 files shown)


rfcs/framework-0040-relations.md
```diff
@@ -0,0 +1,321 @@
...
+```
+{
+    "<field>": {"id": "<persistent id>"},
+    "<field>": [{"id": "<persistent id>"}, ...]
+}
+```
+
+A relation is an object with a single key ``id`` with a string value for a persistent identifier for the related record.
+
+A relation is considered dereferenced if it contains more than one key.
```
fenekku, line 75: I thought the idea was that extensions could be placed in the relation too without necessarily meaning that the relation is dereferenced. e.g.,

undereferenced
```json
{
    "id": "identifierA",
    "ext": {
        "added_field1": "foo",
        "added_field2": "bar"
    }
}

undereferenced

```json
{
    "id": "identifierA",
    "key1": "val1",
    "ext": {
        "added_field1": "foo",
        "added_field2": "bar"
    }
}
```

Granted there could be dereferencing collisions, but that could be solved with a namespaced dereferencing:

new undereferenced

```json
{
    "id": "identifierA",
    "deref": {
        "key1": "val1",
    },
    "ext": {
        "added_field1": "foo",
        "added_field2": "bar"
    }
}
```

rfcs/framework-0040-relations.md
```diff
@@ -0,0 +1,321 @@
...
+
+The syntax was chosen for a number of reasons:
+
+- We would like to avoid confusion with JSONRef as it is already implemented in Invenio, and will work differently than JSONRef.
+- We would like a syntax that can be used both in the internal JSON document stored in the database, as well as in the external JSON used in the REST API.
+- By using an object we can easily dereference it. I.e. ``{"id": "..."}`` is dereferenced to ``{"id": "...", "key1": "val1", ...}``. The key ``id`` always exists (unlike e.g. when using JSONRef).
```
fenekku, line 81:
```suggestion
- By using an object we can easily dereference it i.e. ``{"id": "..."}`` is dereferenced to ``{"id": "...", "key1": "val1", ...}``. The key ``id`` always exists (unlike e.g. when using JSONRef).
```

rfcs/framework-0040-relations.md
```diff
@@ -0,0 +1,321 @@
...
+- Relations should be to a persistent identifier for the related record to make the relation as stable as possible. Also, we wanted to avoid having mulit-key objects as possible persistent identifiers to ensure APIs could be as standardized as much as possible.
+- We did not want to include e.g. type information for the related record, to keep the JSON small as well as keep the type information on the field definition itself, instead of inside the referenced object.
+
+#### Persistent identifiers
+
+We imagine that the persistent identifiers depends on the related record. If the related record uses Invenio-PIDStore, it would be the persistent identifer stored in PIDStore. If the related record just uses the record's table UUID as persistent identifier, this value could be used.
```
fenekku, line 87: Not sure if I follow or if this is too much information: the persistent identifier is always an Elasticsearch record key, right? Whether that key represents the record's table UUID or not doesn't matter, no?

Perhaps a better question is: how does the system know where to go to dereference apart from code? Answer: it doesn't. Only the code knows that.

rfcs/framework-0040-relations.md
```diff
@@ -0,0 +1,321 @@
...
+
+#### Denormalization
+
+Dereferencing a relation can be compared to denormalizing the data or caching the data. This means, that once the related record changes, the places where it was dereferenced/cached must be updated.
+
+The issues with that is:
+
+- **Cardinality**: If a related record is referenced by 1 million records, you have to go and dereference 1 million records again.
+
+- **Concurrency**: With high cardinality, dereferencing will take time, and thus during this time the system can operate with stale data. For instance a new entry may be added, that must be picked up for dereferencing to avoid having it around forever with stale data.
```
fenekku, line 136: What does "that" in "that must be picked up" refer to? 

Isn't it the case that a new entry is the *least* problematic situation? A problematic situation would be:

- Relation to bar is derefenced in a million records
- bar is changed, an asynchronous task A to dereference bar in the million records is launched
- bar is changed again, another asynchronous task B is launched
- task B finishes first on a record and then task A finishes

The asynchronous task should not be carrying the changes themselves to prevent issues. It should just tell the record to dereference again. It will pick up latest changes then.

Not clear to me what are the actual concurrency issues here... don't know that stale data is a problem as long as it is eventually updated (the previous data was fine otherwise we would have had an issue then and if it's now not fine, then there is probably a dependency issue at the root of it really...).

rfcs/framework-0040-relations.md
```diff
@@ -0,0 +1,321 @@
...
+
+##### Database store relations only
+
+Due to the cardinality and concurrency issues, the database should always only store the relation and never a dereferenced version. This avoid that records in the database has to be updated.
+
+##### Version counters
+
+All records have a version counter for optimistic concurrency control. Thus, if everytime we dereference a relation we also store the version counter of the referenced record, we will be able to reliably detect which records must be dereferenced again.
+
+As searches a likely to happen in Elasticsearch, we suggest to make a single compound key
```
fenekku, line 146:
```suggestion
As searches are likely to happen in Elasticsearch, we suggest to make a single compound key.
```

rfcs/framework-0040-relations.md
```diff
@@ -0,0 +1,321 @@
...
+
+We suggest adding a ``@v`` key in the *dereferenced data* (not in the database) that includes the version counter of the referenced record:
+
+```json
+{"id": "1234", "@v": 18, "a": ""}
+```
+
+It should be investigated if this if sufficient in order to proper queries in Elasticsearch. Alternatively, we will need to create a compound key concatenating the id and the version - for instance:
```
fenekku, line 154:
```suggestion
It should be investigated if this if sufficient in order to generate proper queries in Elasticsearch. Alternatively, we will need to create a compound key concatenating the id and the version - for instance:
```

rfcs/framework-0040-relations.md
```diff
@@ -0,0 +1,321 @@
...
+
+
+### Programmatic API
+
+
+#### Defining relations
+
+Relations could be defined by a new system field named ``RelationsField``. The reason for using the name "relations" is to avoid confusion with REST API HATEOS links in the responses.
+
+```pythton
```
fenekku, line 170:
```suggestion
```python
```

rfcs/framework-0040-relations.md
```diff
@@ -0,0 +1,321 @@
...
+```python
+class Record(RecordBase):
+    resource_type = PIDRelation('metadata.resource_type', ResourceTypeRecord.pid)
+```
+
+However, this will make it a bit more diffcult to operate on all relations at once (e.g. ``record.relations.dereference()``).
+
+##### Limiting attributes
+
+Limting the attributes could be done as:
```
fenekku, line 198:
```suggestion
Limiting the attributes could be done as:
```

fenekku (commented):
I had to stop where my last comment is. I will try to finish later if I can.

rfcs/framework-0040-relations.md
lnielsen: The situation is the following:

1. Transaction 1 starts - record Foo v1 is changed to v2
2. Transaction 2 starts - create new record Bar with reference to Foo (v1 is read because v2 is not yet committed)
3. Transaction 1 commits - Foo is now in v2. 
4. Transaction 1 starts a scan of the ES index to find records (same index as Bar) with reference to Foo v1
5. Transaction 2 commits
6. Transaction 2 - record Bar will now reference the relation to Foo. Foo v1 will be picked up, because it's already in the cache.

Result now is:
1) Bar has Foo v1 dereferenced inside the ES index
2) Bar is not included list of records that must be dereferenced again, because the ES index scan is stable and was started before Bar was indexed.

Thus the problem is that the record Bar is never updated and you have forever stale data.

The situation that you describe above is case where the version counter fixes the issue. The referencing task, will always read the latest version, so the only problem there is that we index a million records twice. That's a bit waste of resource, but eventually lead to a consistent search index. The problem above, is that we never reach a consistent search index.

rfcs/framework-0040-relations.md
lnielsen: We went away from it to keep it as simple as possible.

We were looking at solutions of using e.g. ``$id`` instead of ``id`` to know if a record was dereferenced, but then you would end up with that ``record.grant.id`` would not work when the record was not dereferenced (you would have to use ``record.grant.$id`` instead. (e.g. in a jinja template)

rfcs/framework-0040-relations.md
```diff
@@ -0,0 +1,321 @@
...
+```
+
+#### Dereferencing / retrieving the record
+
+```python
+# No relation defined
+record.relations.resource_type() is None
+
+# Getting via metadata
+record['metadata']['resource_type']
```
fenekku, line 212: Add that here the relation is defined:

```suggestion
# Relation defined
## Getting via metadata
record['metadata']['resource_type']
```

rfcs/framework-0040-relations.md
```diff
@@ -0,0 +1,321 @@
...
+# - return only id if  not dereferenced
+# - return id, type and subtype
+
+# Get the related record:
```
fenekku, line 216:
```suggestion
## Getting via relations:
```

rfcs/framework-0040-relations.md
```diff
@@ -0,0 +1,321 @@
...
+```python
+# Via metadata
+Record({
+    "metadata": {
+        "resource_type": {"id": "text-article"}
+    }
+})
+
+# Via system field
+record.relations.resource_type = ResourceType.get_record('text-article')
```
fenekku, line 243: Is the difference between this one and the one below:
- this one sets a dereferenced record
- the one below only sets a non-dereferenced record

Otherwise, when would one use this one versus the one below?

rfcs/framework-0040-relations.md
fenekku: Thanks for the explanations. I think I see the problem. So what is the solution? I see 2 possibilities, (but I am not fluent in the fundamental concurrency tools we have at our disposal) :

1- Out-of-band dereferencing updates: celery beat to dereference all relations of indexed record. (eventually consistent. Could be complementing other approaches too.)
2- Enlarging the optimistic concurrency control scope: Transaction 2 can't commit unless its related records' versions match what is in the database (i.e. if Transaction 1 committed before, then Transaction 2's version for Foo is out of date at time of commit and would be rolled back).

My understanding is that our concurrency tools are limited at ES level, but better at DB level. So perhaps other options at DB...?

rfcs/framework-0040-relations.md
```diff
@@ -0,0 +1,321 @@
...
+
+Updating the outdated references is not doable only via the data access layer.
+
+```python
+# API to scan for records
+service.scan(relations={'resource_type': [('text-article', 18), ...]})
+
+# API to using above scan to reindex
+service.reindex(relations={'resource_type': [('text-article', 18), ...]}, bulk=False)
+```
```
fenekku, line 278: These are just sugar (more convenient syntax), right? We already can search and reindex on ES indices, correct?

fenekku (approved):
Thanks for the explanations. I listed more questions (concurrency again), but the overall approach seems fine to me.

PR #45, opened 2021-03-09
rdm: versioning support

lnielsen:


rfcs/rdm-0045-versioning-support.md
```diff
@@ -0,0 +1,530 @@
...
+- ``version_id`` - Version counter used for optimistic concurrency control.
+- ``created `` - Creation timestamp.
+- ``updated`` - Modification timestamp.
+- ``bucket_id`` - Foreign key (UUID) to a Files-REST bucket.
+
+We propose to add the following two attributes:
+
+- ``parent_id`` - Foreign key (UUID) to a Concept Record (a new entity).
+- ``parent_order`` - Integer (4 bytes) defining the order of siblings.
+- ``parent_latest`` - Boolean to indicate if the record is the latest version.
```
fenekku, line 138: I know commits have already been merged, but here are my takes on the names:

``parent_id`` is both the id of the parent *and* namespacing for the "versioning" features. This makes ``parent_order`` and ``parent_latest`` confusing because ``parent`` there is only for namespacing. How about: 

```suggestion
- ``versioning_parent_id`` / ``versioning_concept_id`` - Foreign key (UUID) to a Concept Record (a new entity).
- ``versioning_order`` - Integer (4 bytes) defining the rank of the record among sibling versions.
- ``versioning_latest`` - Boolean to indicate if the record is the latest version.
```

rfcs/rdm-0045-versioning-support.md
```diff
@@ -0,0 +1,530 @@
...
+
+In addition a database index should be defined to allow for easily querying all records related to a concept record in the correct order (thus likely a combined index on ``parent_id`` and ``parent_order``):
+
+```SQL
+SELECT * FROM rdm_records_metadata WHERE parent_id=... ORDER BY parent_order
+```
+
+#### Concept Record
+
+We also propose to add a new entity, a ``ConceptRecord``, which represents the concept record and are able to hold properties for all records such as ownership and permissions.
```
fenekku, line 150:
```suggestion
We also propose to add a new entity, a ``ConceptRecord``, which represents the concept record and is able to hold properties for all records such as ownership and permissions.
```

rfcs/rdm-0045-versioning-support.md
```diff
@@ -0,0 +1,530 @@
...
+}
+```
+
+Essentially the concept persistent identifier, the access ownership and access grants would be moved from the existing record to the new concept record.
+
+**Why a new entity?**
+
+The new model allows keeping the ownership and permissions separate from the single version records. Alternatively, we would have to duplicate (and update) access information in multiple specific version records.
+
+Also, having just a single concept record for both records and drafts, means that access information can be updated not only from the deposit form, but also e.g. share button on the landing page record.
```
fenekku, line 183: (it's really the fact that drafts and recors share the same recid that makes it easy to have share button on each :wink: ). Having a single concept record for both record and drafts is "natural":

```
Concept Record
|
|___ Record V2
|___ Record V1 ------ Draft V1
|___ Record V0 
```

rfcs/rdm-0045-versioning-support.md
```diff
@@ -0,0 +1,530 @@
...
+
+In addition, it opens future use cases where we may want to associate more data with the concept record. On possible upcoming use case involves storing shared links with the record.
+
+We will also know exactly when we have to reindex siblings and when it's not needed.
+
+**Why not in the JSON?**
+
+An alternative is to store the ``parent_id`` and ``parent_order`` in the record instead of at the database table, however would like to have the database-level constraints enforced.
```
fenekku, line 191:
```suggestion
An alternative is to store the ``parent_id`` and ``parent_order`` in the record instead of at the database table, however we would like to have the database-level constraints enforced and take advantage of indexing on these fields.
```

rfcs/rdm-0045-versioning-support.md
```diff
@@ -0,0 +1,530 @@
...
+
+```python
+{
+    "concept": {
+        "id": "<uuid>",
+        "access": {
+            "owned_by
+        }
+    },
+}
```
fenekku, line 211: So we don't try to hide the technical level implementation at the API level (i.e. put all access in same place and keep `conceptid` as before)?  We should perhaps then be explicit:

```suggestion
}

The `conceptid` field is removed from the API in favour of the above.
```

rfcs/rdm-0045-versioning-support.md
```diff
@@ -0,0 +1,530 @@
...
+service.edit(id_, identity, data, links_config=...)
+```
+
+#### Create a new version
+
+```python
+service.new_version(id_, identity, data, links_config=...)
+```
+
+#### Updating a draft
```
fenekku, line 253:
```suggestion
#### Update a draft

service.update_draft(id_, identity, data. links_config=...)
```

rfcs/rdm-0045-versioning-support.md
```diff
@@ -0,0 +1,530 @@
...
+
+The update of a draft is unchanged, except that you can no longer change the ownership and permissions via the draft (which anyway was not possible in the February release).
+
+**Visibility and embargo information**
+
+Specifically, you will still update the record/files restrictions and embargo date/reason via the is method. The only difference is that it requires manage permissions to change this information, where as the metadata can be changed with edit permission.
```
fenekku, line 259:
```suggestion
Specifically, you will still update the record/files restrictions and embargo date/reason via this method (granted you have the `can_edit` permission). 
```

rfcs/rdm-0045-versioning-support.md
```diff
@@ -0,0 +1,530 @@
...
+
+#### Updating ownership and permissions
+
+We propose adding a new method specifically update the permissions:
```
fenekku, line 263:
```suggestion
We propose adding a new method to specifically update the ownership and permissions:
```

rfcs/rdm-0045-versioning-support.md
```diff
@@ -0,0 +1,530 @@
...
+
+```python
+service.update_permissions(id_, identity, data, links_config=...)
+service.update_owners(id_, identity, data, links_config=...)
+```
+
```
fenekku, line 269: Move the bit above here:

```suggestion
This method requires `can_manage` permission to change this information.
```

rfcs/rdm-0045-versioning-support.md
```diff
@@ -0,0 +1,530 @@
...
+The existing REST API, already define endpoints for the main operations discussed in the service:
+
+- ``GET /records`` - Search records.
+- ``GET /user/records`` - Search drafts.
+- ``POST /records`` - Create a new draft.
+- ``POST /records/:id/draft`` - Edit a record.
+- ``PUT /records/:id/draft`` - Update draft.
+- ``POST /records/:id/actions/publish`` - Publish a record.
+- ``GET /record/:id/versions`` - Search versions for a record.
+- ``POST /record/:id/versions`` - Create a new version.
```
fenekku, line 307:
```suggestion
- ``GET /records/:id/versions`` - Search versions for a record.
- ``POST /records/:id/versions`` - Create a new version.
```
right?

rfcs/rdm-0045-versioning-support.md
```diff
@@ -0,0 +1,530 @@
...
+
+We will add a new endpoint for updating the permissions:
+
+- ``PUT /records/:id/access`` - Update permissions.
+
+We use ``PUT`` instead of e.g. ``PATCH /records/:id`` because we want to make use of the concept records version counter in the ETag for optimistic concurrency control, and we want to model the endpoint as a new REST resource. Some notes:
+
+- Identifier: We're using the ``:id`` of single version records instead of the concept record. This is to make it easier for API developers to deal only with a single identifier for a version, and to not confusing by using a different id.
```
fenekku, line 315:
```suggestion
- Identifier: We're using the ``:id`` of single version records instead of the concept record. This is to make it easier for API developers: they only have to deal with a single identifier (the one for a record version), they don't have to make an additional API call, and they don't need to track another identifier.
```

rfcs/rdm-0045-versioning-support.md
```diff
@@ -0,0 +1,530 @@
...
+- ``POST /records`` - Create a new draft.
+- ``POST /records/:id/draft`` - Edit a record.
+- ``PUT /records/:id/draft`` - Update draft.
+- ``POST /records/:id/actions/publish`` - Publish a record.
+- ``GET /record/:id/versions`` - Search versions for a record.
+- ``POST /record/:id/versions`` - Create a new version.
+
+We will add a new endpoint for updating the permissions:
+
+- ``PUT /records/:id/access`` - Update permissions.
```
fenekku, line 311:
```suggestion
We will add new endpoints:

- ``PUT /records/:id/access`` - Update permissions.
- ``GET /records/:id/versions/latest - Get the latest version
```

according to `links` below.

rfcs/rdm-0045-versioning-support.md
```diff
@@ -0,0 +1,530 @@
...
+```json
+{
+    "id": "...",
+    "access": {
+        "owned_by": []
+        "grants": []
+    }
+    "links": {
+        "access": "https://localhost:5000/api/records/:id/access"
+    }
```
fenekku, line 346: this is itself right? Are we returning self-referential links for the files metadata endpoint? This is the endpoint that is the most similar. For records we call these links `self`, but is this really needed in this case...?

rfcs/rdm-0045-versioning-support.md
```diff
@@ -0,0 +1,530 @@
...
+- Newer version alert
+- Previous versions
+- Manage section:
+    - A "New version" button.
+
+**Newer version alert**
+
+A record that has a newer version should clear display a warning message (same as on Zenodo), with a link to the new version.
+
+The condition for the new version is essentially: ``record.parent_latest is False``.
```
fenekku, line 375: :+1:

rfcs/rdm-0045-versioning-support.md
```diff
@@ -0,0 +1,530 @@
...
+**Version facet**
+
+A flip-switch which essentially
+
+- Version facet (flip switch): ``?versions=all|latest``
+- Sort by version number
+- Filter by parent version
+- Result item: display link to concept landing page
+
+## Plan
```
fenekku, line 410: assuming from here on out it's just scratch pad.

fenekku (approved):
Thanks for this. Spent a good chunk of my day reading it through and understanding (at some high-level) what is going on with versioning.

PR #50, opened 2021-09-02
Add RFC for data citation recommendations

mfenner:
This pull request adds a high-level overview of the planned work to support the [Data Citation Recommendations](https://doi.org/10.1038/s41597-019-0031-8) in invenioRDM. Most of the planned work is straightforward, but some topics need discussion, in particular:

* full implementation of `givenName` and `familyName` for personal authors. This is needed to properly expose author metadata in `schemaOrg` and `bibtex`, but is a breaking change.
* scope and detail of documentation for data citation that needs to be added
* whether or not to implement the optional recommendations **content negotation** and **http link headers**.


mfenner:
Updated PR after call with @lnielsen on September 6.

PR #56, opened 2022-02-01
rdm: events bus

ppanero:
edit link: https://codimd.web.cern.ch/X1PJE3dLRvibdCnHznPxsg

### DIFF
--- a/rfcs/rdm-0056-event-bus.md
+++ b/rfcs/rdm-0056-event-bus.md
@@ -1,0 +1,311 @@
+- Start Date: 2022-02-01
+- RFC PR: [#56](https://github.com/inveniosoftware/rfcs/pull/56)
+- Authors: Pablo Panero, Alex Ioannidis  
+
+
+# Event Bus
+
+## Summary
+
+This RFC defines the implementation design of the _event bus_ pattern in order to provide support for event-driven design in InvenioRDM.
+
+## Motivation
+
+The _Unit of Work_ pattern has been implemented in Invenio in order to group several actions into one atomic operation, meaning one database commit.
+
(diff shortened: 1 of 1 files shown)


rfcs/rdm-0056-event-bus.md
```diff
@@ -0,0 +1,311 @@
...
+Two alternatives were considered to the event worker task are:
+
+- Delegating the handling to the event bus itself. However, this would put even more responsability on `invenio-rdm-records`, which is not desired at the moment.
+- Having a new app run with for example a `supervisord`. This would increase the maintainance cost significantly to provide scalability, resilience, etc.
+
+## Unresolved questions
+
+**How to add events that require request context**
+
+Is this suitable for cases where request context is required? For example usage statistics would require IP addr, referer, User Agent among others. These are available at _resource_ level, how can an event be added/queued from there?
```
ntarocco, line 292: This should not be a problem at all: anything that is needed in the event should be passed from the resource layer to the service layer as args, and stored it in specific fields of the `dataclass`. The event dataclass should be self-contained.

rfcs/rdm-0056-event-bus.md
```diff
@@ -0,0 +1,311 @@
...
+
+**How to deal with cases that do not require a UoW**
+
+For example, record `read` operations do not require a _unit of work_ and therefore it is not possible add events such as `RecordReadEvent`.
```
ntarocco, line 296: If not mistaken, we can still add the `uow` to simply send an event and nothing else.

rfcs/rdm-0056-event-bus.md
```diff
@@ -0,0 +1,311 @@
...
+
+**Error handling**
+
+Error handling plays a bit role in this event bus implementation. While there are mechanisms to potentially allow persistance of events there are no mechanisms to:
```
ntarocco, line 300: This is a very good point: the event success should be recorded only when the handler has completed the job. It is probably the part that should be expanded in this RFC.

slint:
Merged for archival/history purposes

PR #57, opened 2022-02-17
Framework: Declarative Python package definitions

lnielsen:


rfcs/framework-0057-declarative-python-package-definition.md
```diff
@@ -0,0 +1,174 @@
...
+
+Having a minimal ``setup.py`` as below allow us have almost the entire package work as previously:
+
+```python
+# setup.py
+from setuptools import setup
+setup()
+```
+
+This way tools like Requirements Builder work without modifications and you can still install a package via ``python setup.py install`` and build docs and compile message catalogs via ``python setup.py build_spinhx/compile_catalog``.
```
fenekku, line 49:
```suggestion
This way tools like Requirements Builder work without modifications and you can still install a package via ``python setup.py install`` and build docs and compile message catalogs via ``python setup.py build_sphinx/compile_catalog``.
```

rfcs/framework-0057-declarative-python-package-definition.md
```diff
@@ -0,0 +1,174 @@
...
+- License: The license classifier is already specified in the ``license`` keyword.
+- Python versions: The version classifiers were not kept updated, and where purely metadata, whereas the ``python_requires`` keyword is actionable and easier to keep updated.
+
+#### Version number
+
+The version number has to be moved from ``<package>/version.py`` to ``<package>/__init__.py``. This is because setuptools is only able to extract the version number using the Abstract Syntax Tree (i.e. not executing the code) when located in the ``__init__.py``. If the version number is placed in ``version.py`` setuptools wih import the main package which may lead to imports failing.
+
+#### Extras require
+
+We can no longer programmatically define an ``all`` extra from other extras (like ``tests`` and ``docs``. There's two options:
```
fenekku, line 128:
```suggestion
We can no longer programmatically define an ``all`` extra from other extras (like ``tests`` and ``docs``. There are two options:
```

rfcs/framework-0057-declarative-python-package-definition.md
```diff
@@ -0,0 +1,174 @@
...
+
+- Duplicate all extra dependencies in an ``all`` extras (risk of getting outdated).
+- Gather as much as possible into a ``tests`` extras (extra documentation needed).
+
+We suggest as far as possible keeping a single ``all`` extra requires that includes tests and docs requirements. Extra requirements needs to be duplciated into this all requirements, but this means that all packages can mostly be installed via ``pip install -e ``.[all]`` to start development on the package.
```
fenekku, line 133:
```suggestion
We suggest as far as possible keeping a single ``all`` extra requires that includes tests and docs requirements. Extra requirements needs to be duplicated into this ``all`` requirements, but this means that all packages can mostly be installed via ``pip install -e ``.[all]`` to start development on the package.
```

This is option 1 right? My vote is for that one too: the duplication is for records listed right next to each other so them getting out-of-sync is much less likely.

rfcs/framework-0057-declarative-python-package-definition.md
```diff
@@ -0,0 +1,174 @@
...
+No changes neeed.
+
+## How we teach this
+
+Above should be implemented in the ``cookiecutter-invenio-module`` template repository so that it can always be used as a reference implementation of a Python package for Invenio.
+
+
+## Drawbacks
+
+The benefits of this change only becomes available on the long-term as there are no immediate benefits. However long-term it enables us to improve and automate our dependency management that continue to be a major source of disruptions to the development.
```
fenekku, line 170:
```suggestion
The benefits of this change only becomes available on the long-term as there are no immediate benefits. However long-term it enables us to improve and automate our dependency management which continues to be a major source of disruptions to development.
```

fenekku (approved):
:heart_decoration:

PR #62, opened 2022-04-14
framework: change notifications

ppanero:


ppanero:
@slint this is implemented and actually being used since v9. Maybe it would be good to have it merged before we go into the message bus?

slint:
Yup, let's merge it as is, and mark as implemented.

PR #66, opened 2022-09-01
rfc: notification-system

### LINKED ISSUE #63: [Proposal] (E-Mail) Notification System
[Codimd](https://codimd.web.cern.ch/RH_steApSGOk4Yaoex782A) for RFC.  

#### Use Cases
- Send notifications (see [table](#notifications))
- Overridable notification templates
  - One template for each notification
  - Can be overridden via theme (like other templates)
  - Add new templates for custom notifications
  - Should be translatable
    - Store preferred language in user profile
- Modify notifications
  - Set which notifications should be send
    - As a repository admin activate/deactive certain notifications
    - As a user (user profile)
  - Define events per module
  - Define notification backends per module
- Configure
  - Define notification config per module
    - Backends used
    - Events happening
  - Combine in `app-rdm`

#### Notifications
| Action                       | Who triggers                 | Who receives           |
|------------------------------|------------------------------|------------------------|
| Comment create      | Creating user    | All involved in conversation                  |
| Comment edit        | Editing user     | All involved in conversation                  |
| Comment delete      | Deleting user    | All involved in conversation                  |
|       |     |                  |
| Submission create      | Creating user     | Community owners/managers                 |
| Submission accept      | Accepting user    | User who created submission               |
| Submission decline     | Declining user    | User who created submission               |
| Submission edit        | Editing user      | Either user who created submission or Community owners/managers (depends on who edited)                 |
|       |     |                  |
| Invitation accept      | Accepting user    | User who created invitation               |
| Invitation decline     | Declining user    | User who created invitation               |
| Invitation expire      | System            | User who created invitation               |

rekt-hard:
- closes #63 
- closes https://github.com/inveniosoftware/invenio-notifications/issues/4

### DIFF
--- a/rfcs/framework-0063-notification-system.md
+++ b/rfcs/framework-0063-notification-system.md
@@ -13,10 +13,10 @@ Implementation details for a notification system to be used in InvenioRDM.
 As the product is getting larger and new features are added, it is hard to keep an overview of everything going on as a user. With the communities feature, certain tasks require an action to be executed by the user (f.e. accepting an incoming request for a community). This incoming request can easily get "lost", if a user does not explicitly check the community dashboard. Therefore, a notification system will not only solve this problem (and similar ones) but also increase user engagement by advertising the repository some more.
 
 It will take care of sending the notifications based on the configuration and user preferences (i.e. if the repository disables certain notifications or a user does not want to receive certain notifications, they should not be sent).
-In a first iteration, the user preferences may be neglected. They have to be specified and implemented first or the recipients function will have to take care of it explicitly.
 
-### Use Cases
-- Send notifications (see [table](#notifications))
+### Use Case
+
+- Send notifications for operations involving multiple users (see [table](#notification-examples))
 - Overridable notification templates
   - One template for each notification
   - Can be overridden via theme (like other templates)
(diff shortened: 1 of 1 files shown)


rfcs/framework-0063-notification-system.md
```diff
@@ -0,0 +1,422 @@
...
+
+# Notification System
+
+## Summary
+
+Implementation details for a notification system to be used in InvenioRDM.
+
+## Motivation
+
+As the product is getting larger and new features are added, it is hard to keep an overview of everything going on as a user. With the communities feature, certain tasks require an action to be executed by the user (f.e. accepting an incoming request for a community). This incoming request can easily get "lost", if a user does not explicitly check the community dashboard. Therefore, a notification system will not only solve this problem (and similar ones) but also increase user engagement by advertising the repository some more.
```
ppanero, line 13:
```suggestion
As the product is getting larger and new features are added, it is hard to keep an overview of everything going on as a user. With the communities feature, certain tasks require an action to be executed by the user (f.e. accepting an inclusion request for a community). This inclusion request can easily be "missed", if a user does not explicitly check the community dashboard. Therefore, a notification system will not only solve this problem (and similar ones) but also increase user engagement by advertising the repository some more.
```

rfcs/framework-0063-notification-system.md
```diff
@@ -0,0 +1,422 @@
...
+
+It will take care of sending the notifications based on the configuration and user preferences (i.e. if the repository disables certain notifications or a user does not want to receive certain notifications, they should not be sent).
```
ppanero, line 15:
```suggestion
The system will take care of sending the notifications based on the configuration and user preferences (i.e. if the repository disables certain notifications or a user does not want to receive certain notifications, they should not be sent).
```

rfcs/framework-0063-notification-system.md
```diff
@@ -0,0 +1,422 @@
...
+
+### Use Case
```
ppanero, line 17:
```suggestion
### Use Cases
```

rfcs/framework-0063-notification-system.md
```diff
@@ -0,0 +1,422 @@
...
+
+- Send notifications for operations involving multiple users (see [table](#notification-examples))
```
ppanero, line 19: Many points of the enumeration are hard to grasp and the implications are mixed (e.g. at a user profile level, at a template or instance level, etc.). Maybe using the templated [as a ... I want ... so that ...](https://github.com/inveniosoftware/rfcs/blob/master/0000-template.md#motivation) structure would clarify the use cases

rfcs/framework-0063-notification-system.md
```diff
@@ -0,0 +1,422 @@
...
+In this design, we describe:
+
+- New `invenio-notifications` module
+    - Simple self-contained API that is business-logic agnostic
+    - Minimal dependencies, only for supporting different notification transfer backends (email, chat, etc.)
+    - Defines a common notification payload/datamodel
+- Utilities for configuration on services for building notifications payloads (receivers, data payload)
+    - Generates the list of recipients
+    - Takes into account user preferences, e.g.
+        - if they want to receive notifications at all
```
ppanero, line 44:
```suggestion
        - If they want to receive notifications at all
```

rfcs/framework-0063-notification-system.md
```diff
@@ -0,0 +1,422 @@
...
+        - what kind of notifications the want to receive
```
ppanero, line 45:
```suggestion
        - What kind of notifications the want to receive
```

rfcs/framework-0063-notification-system.md
```diff
@@ -0,0 +1,422 @@
...
+    # Options/parameters for the backend to deliver the notification
+    params: dict
+```
+
+### API
+
+A Notification manager/client will be used for sending notifications as a Celery task.
+
+```python
+class NotificationManager:
```
ppanero, line 102: Higher level question: Not a fan of `manager`, although we have it in other places too (e.g. pids). I am assuming they are at service level, do we have a clear scope of what they should and should not do? otherwise they might become a bit of a `catch-all` methods that do not have a clear place. I just feel when we do not find a place we just create a "manager".

The API for this specific manager makes sense 👌

rfcs/framework-0063-notification-system.md
```diff
@@ -0,0 +1,422 @@
...
+
+    def __init__(self, backends):
+        self.backends = backends
+
+    def validate(self, notification):
```
ppanero, line 107: would validate raise exceptions or return `True` or `False`? this would affect the `broadcast` function below.

rfcs/framework-0063-notification-system.md
```diff
@@ -0,0 +1,422 @@
...
+        ...
+
+    # Client
+    def broadcast(self, notification: Notification, eager=False):
```
ppanero, line 111: Does this mean we can start typing params and return values 😍 ?

rfcs/framework-0063-notification-system.md
```diff
@@ -0,0 +1,422 @@
...
+        """Broadcast a notification via a Celery task."""
+        self.validate(notification)
+        task = broadcast_notification.si(notification)
```
ppanero, line 114: ?
```suggestion
        task = broadcast.si(notification)
```

rfcs/framework-0063-notification-system.md
```diff
@@ -0,0 +1,422 @@
...
+
+    # Consumer
+    def handle(self, notification: Notification):
+        """Handle a notification."""
+        for recipient in notification.recipients:
+            for backend in recipient.backends:
+                dispatch_notification(backend, recipient.data, notification.data)
+```
+
+The `broadcast_notification` and `dispatch_notification` tasks will be handling retries:
```
ppanero, line 125:
```suggestion
The `broadcast` and `notify` tasks will be handling retries:
```

the two below right? although then I am a bit confused the might be a loop between manager.broadcast and the broadcast task?

rfcs/framework-0063-notification-system.md
```diff
@@ -0,0 +1,422 @@
...
+    """Broadcast notification to backends according to event policy."""
+    current_notifications_manager.broadcast(notification)
+
+@shared_task
+def notify(notification, backend_id):
+    """Set notification and notify specific backend.
+
+    Will pass the key for the specific backend to notify.
+    """
+    current_notifications_manager.notify(notification, backend_id)
```
ppanero, line 140: where is `notify`? it it handle?

rfcs/framework-0063-notification-system.md
```diff
@@ -0,0 +1,422 @@
...
+from abc import ABC, abstractmethod
+
+class Backend(ABC):
+
+    id = None
+    """Unique ID of the backend."""
+
+    @abstractmethod
+    def send(self, notification, recipient, params):
+        """Here each concrete implementation will be able to consume the notification message."""
```
ppanero, line 166:
```suggestion
        """Consume the notification message."""
```

rfcs/framework-0063-notification-system.md
```diff
@@ -0,0 +1,422 @@
...
+    backends = [
+        UserEmailBackend,
+    ]
+
+#
+# Payload
+#
+# invenio_requests/notifications.py
+class RequestPayload:
+    def __call__(self, request=None, **kwargs):
```
ppanero, line 286: overwriting call in this case would mean having things like `RequestPayload()()` which is not very intuitive. Is there a reason to have `__call__` overwriten instead of having a method return this values? I kind of like the `__call__` but also makes it not clear in interface for others to create a Payload class (is there an interface?).

rfcs/framework-0063-notification-system.md
```diff
@@ -0,0 +1,422 @@
...
+            user = rec.data
+            rec.backends.append({
+                "backend": "email",
+                "to": f"{user.profile.full_name} <{user.email}>",
+            })
+
+#
+# Usage in a service method
+#
+uow.register(NotificationOp(CommunityRecordInclusionNotificationBuilder.build(
```
ppanero, line 372: Where would `NotificationOp` process things, `post_commit`? would it be worth to add it to the RFC?

rfcs/framework-0063-notification-system.md
kpsherva: does it make sense to mention a default behaviour? f.e. by default, the user receives all the notifications.

rfcs/framework-0063-notification-system.md
kpsherva: plus one to this :)

rfcs/framework-0063-notification-system.md
kpsherva, line 45:
```suggestion
        - what kind of notifications they want to receive
```

rfcs/framework-0063-notification-system.md
kpsherva: I think I am a fan of more elaborate naming, when you use it outside the module, it will be explicit (unless we are not planning to expose it to the client's use)

rfcs/framework-0063-notification-system.md
```diff
@@ -0,0 +1,422 @@
...
+```python
+from invenio_notifications import current_notifications
+
+notification = ...  # see "Building notifications"
+current_notifications.broadcast(notification)
+```
+
+### Backends
+
+A notification backend will receive all the information from the notification system through a notification (i.e. type of action/event, user who triggered,  recipients, community/record (based on action/event)) to construct a suitable notification. The backend itself should *not* have to perform any queries (ElasticSearch or database). Its only purpose should be to take the received information, create the payload for the notification and send it to the recipients.
```
kpsherva, line 154: what about:
```suggestion
A notification backend will receive all the information from the notification system through a notification payload (i.e. type of action/event, user who triggered,  recipients, community/record (based on action/event)) to construct a suitable notification.
The backend itself should *not* have to perform any queries (ElasticSearch or database). Its only purpose should be to take the received information, create the payload for the notification and send it to the recipients.
```

rfcs/framework-0063-notification-system.md
```diff
@@ -0,0 +1,422 @@
...
+
+> How should this feature be introduced and taught to existing Invenio users?
+
+## Alternatives
+
+An alternative is to rely on the implementation of the event bus and register methods. Though it is not known which data will be passed along for the events and there may be some missing data for the notification system.
+
+## Unresolved questions
+
+Tracking success/failure of delivery is not taken care of. If a task gets stuck or dies, there will not be any mention of this.
```
kpsherva, line 418: silly question, are we use rabbitmq for messages in this case? do notifications get their own queues?

rfcs/framework-0063-notification-system.md
slint: Agree, these could be just 3 functions tbh in the module, since at the end of the day we're just using them like so...

One of the older suggestions was to just have `Notifications` as a class name. 

I think we can leave this up to the actual implementation for now, since this RFC is already pretty packed with code blocks.

rfcs/framework-0063-notification-system.md
slint: This is meant to be called by the service and validates the produced notification payload, e.g. if required backend parameters are there (`email` always needs a `to` field, etc.).

The issue is that it's called in the `on_post_commit` clause of a UoW operation, so it should have the same behavior as e.g. failure of ElasticSearch indexing or sending a Celery task... In REST API this is a 500 response :)

rfcs/framework-0063-notification-system.md
slint: These have been renamed now a couple of times, so I think I forgot to update their names, and that's why it's not very clear which task does what. Updating now to something more specific.

rfcs/framework-0063-notification-system.md
ppanero: Guess: yes. At the moment we have one queue per record type (e.g. funders, awards, names, etc.) so I guess notifs will have their own (or even more fine grained). This would also allow to set priorities to the queues (🤔), useful?

rfcs/framework-0063-notification-system.md
```diff
@@ -0,0 +1,422 @@
...
+| ------------------ | -------------- | --------------------------------------------------------------------------------------- |
+| Comment create     | Creating user  | All involved in conversation                                                            |
+| Comment edit       | Editing user   | All involved in conversation                                                            |
+| Comment delete     | Deleting user  | All involved in conversation                                                            |
+|                    |                |                                                                                         |
+| Submission create  | Creating user  | Community owners/managers                                                               |
+| Submission accept  | Accepting user | User who created submission                                                             |
+| Submission decline | Declining user | User who created submission                                                             |
+| Submission edit    | Editing user   | Either user who created submission or Community owners/managers (depends on who edited) |
```
lnielsen, line 392: I don't think we're able to detect this event. We should log an event on the request timeline when a user edits an associated record (i.e. when in submitted state). But we don't want 20 events logged when a user edit the record and press save.

rfcs/framework-0063-notification-system.md
```diff
@@ -0,0 +1,422 @@
...
+  - Should be translatable
+    - Store preferred language in user profile
+- Modify notifications
+  - Set which notifications should be send
+    - As a repository admin activate/deactive certain notifications
+    - As a user (user profile)
+  - Define notifications for events per module
+  - Define notification backends per module
+
+## Detailed design
```
lnielsen, line 33: I'm missing statements about which layer which parts belong to in the software architecture: https://inveniordm.docs.cern.ch/develop/architecture/software/

rfcs/framework-0063-notification-system.md
```diff
@@ -0,0 +1,422 @@
...
+  - Can be overridden via theme (like other templates)
+  - Add new templates for custom notifications
+  - Should be translatable
+    - Store preferred language in user profile
+- Modify notifications
+  - Set which notifications should be send
+    - As a repository admin activate/deactive certain notifications
+    - As a user (user profile)
+  - Define notifications for events per module
+  - Define notification backends per module
```
lnielsen, line 31: How is it going to look for like a user? 

I'd like to see
- Profile page: Disable/enable notification backends, edit preferred language
- Examples of possible notifications sent by e.g. email and say Mattermost/Slack? E.g. the email should likely have information about how to "unsubscribe" - i.e. change their notification settings. On Zenodo we may want to ask users wan to receive informational emails so we're able to send messages to all users that want to receive them.

Let me know if you'd like me to help with mockups

ntarocco: After chat, what is probably missing here is to describe with more details that the templates will be also per-backend.
And yes, the plan was to have e-mail notifications, and immediately add `CERN Mattermost` as extra to test that it actually works adding another backend.

rfcs/framework-0063-notification-system.md
```diff
@@ -0,0 +1,422 @@
...
+    }
+)))
+```
+
+#### Notification examples
+
+| Action             | Who triggers   | Who receives                                                                            |
+| ------------------ | -------------- | --------------------------------------------------------------------------------------- |
+| Comment create     | Creating user  | All involved in conversation                                                            |
+| Comment edit       | Editing user   | All involved in conversation                                                            |
```
ntarocco, line 386: This might be overkill. I don't think that GitHub sends out notifications when a comment is updated.

rfcs/framework-0063-notification-system.md
```diff
@@ -0,0 +1,422 @@
...
+| Comment delete     | Deleting user  | All involved in conversation                                                            |
```
ntarocco, line 387: Comment edit/delete might be overkill. I don't think that GitHub sends out notifications when a comment is updated or deleted.

rfcs/framework-0063-notification-system.md
```diff
@@ -0,0 +1,422 @@
...
+|                    |                |                                                                                         |
+| Submission create  | Creating user  | Community owners/managers                                                               |
+| Submission accept  | Accepting user | User who created submission                                                             |
```
ntarocco, line 390:
```suggestion
| Submission accept  | Accepting user | User who created submission, only when not the same user |
```

rfcs/framework-0063-notification-system.md
slint: These should be clear now in the RFC, the manager class has the following high-level methods:

- `broadcast(notification)`: called by the "client", i.e. the service layer that triggers the notification. It does some basic validation and sends a Celery task with a minimal notification payload
- `handle_broadcast(notification)`: called by the Celery task from `broadcast(...)` above. Responsible for:
  - expanding the notification context via "read" calls to the service layer
  - building the list of recipients
  - filtering the list based on recipient preferences
  - figuring out what backend each recipient wants
  - sending of a Celery task for the actual delivery for each recipient backend
- `handle_dispatch(backend, recipient, notification)`: called by the Celery task for `handle_broadcast(...)` above. Responsible for calling the backend to deliver the message to a recipient

rfcs/framework-0063-notification-system.md
slint: The pseudo-implementation for these has been updated (with some changed semantics as well of when they run). There's a more generic set of classes now, receiving a path/key for an entity reference to resolve. So the final usage looks something like:

```python
# usage in the NotificationBuilder class:
context = [
    EntityResolve(key="request"),
    EntityResolve(key="request.created_by"),
    EntityResolve(key="request.topic"),
    EntityResolve(key="request.receiver"),
]   

class EntityResolve:
    def __init__(self, key):
        self.key = key

    def __call__(self, notification: Notification):
        entity_ref = dict_lookup(notification, self.key)
        entity = EntityResolverRegistry.resolve_entity(entity_ref)
        dict_set(notification, self.key, entity)
        return notification
```

The idea is also to design the interface around simple `callable`s instead of some arbitrary `.{run,process,resolve}(...)` method.

rfcs/framework-0063-notification-system.md
slint: The `NotificationOp` is not doing any processing, but delegates to a `NotificationBuilder` creating the notification payload/context containing just entity references to the underlying objects. The "heavy" processing has been updated to take place in the `NotificationManager` class in a Celery task.

rfcs/framework-0063-notification-system.md
```diff
@@ -0,0 +1,422 @@
...
+    - As a user (user profile)
+  - Define notifications for events per module
+  - Define notification backends per module
+
+## Detailed design
+
+In this design, we describe:
+
+- New `invenio-notifications` module
+    - Simple self-contained API that is business-logic agnostic
```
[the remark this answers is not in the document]
slint: RFC has been updated accordingly with correct responsibilities.

rfcs/framework-0063-notification-system.md
slint: For now, there's no real benefit in creating separate queues for the notification message transmission itself. Using separate queues would also go in the direction of events/message bus, which we're trying to avoid to reduce complexity.

The plan is to just use Celery tasks at the moment that get shipped to the primary/default queue. This can be configured of course, but currently in RDM we don't have multiple Celery queues with different priorities, I to reduce deployment/infrastructure complexity.

rfcs/framework-0063-notification-system.md
```diff
@@ -0,0 +1,422 @@
...
+    def __call__(self, notification, recipients: list):
+        ret = []
+        if self.include_creator:
+            ret.append(request.created_by.resolve())
+        if self.include_receiver:
+            ret.append(request.receiver.resolve())
+        return ret
+
+# invenio_communities/notifications.py
+class CommunityRecipients:
```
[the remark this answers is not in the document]
slint: Yes, module-specific payload and recipient resolution logic will live in the respective modules. This is also to bind the logic with the datamodel/schema, e.g. `invenio-users-resources` defines the "global notification on/off switch" schema, so it provides the utility to filter a recipient based on this.

rfcs/framework-0063-notification-system.md
```diff
@@ -0,0 +1,422 @@
...
+    def __init__(self, roles=None):
+        self.roles = roles
+
+    def __call___(self, notification, recipients: list):
+        ret = []
+        for rec in recipients:
+            if isinstance(rec, Community):
+                comm = rec
+                members = Member.get_members(comm.id)
```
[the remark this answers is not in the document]
slint: RFC has been changed to make it clear that notifications are Presentation layer and that context/payload and recipient resolvers are fetched using the Service layer.

rfcs/framework-0063-notification-system.md
```diff
@@ -0,0 +1,422 @@
...
+    - Generates the list of recipients
+    - Takes into account user preferences, e.g.
+        - if they want to receive notifications at all
+        - what kind of notifications the want to receive
+
+### Datamodel/payload
+
+- Acts as a "client payload"
+- Keeps all the relevant data necessary for a notification to be dispatched via a backend
+- It must be JSON-serializeable so that it can be passed to Celery tasks.
```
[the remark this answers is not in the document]
slint: Payload has been updated in the RFC to be minimal right now (just an entity reference to the request or request event).

rfcs/framework-0063-notification-system.md
slint: > How is it going to look for like a user?

Mockups at https://github.com/inveniosoftware/invenio-communities/issues/864#issuecomment-1497107633. But not part of the initial implementation because of time-constraints and planning requirements. Will add to the RFC though for reference.

We're still missing mockups for the rendered email/chat templates, in the current implementation we'll just have the (hyper)text message. These can be mocked-up and styled later on as we did e.g. for registration, and email confirmation emails.

> Question: If we wanted to build a page listing notifications for a user, would this module be able to support it? i..e would it just be a matter of a new backend storing the notifications?

Yes, we can just add a backend that is used to persist the message to primary/secondary storage. Probably this can live in this module and have it's own service, resources, UI in the user settings page, etc.

rfcs/framework-0063-notification-system.md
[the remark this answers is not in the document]
slint: This runs in a Celery task now, so we're safe in such cases.

rfcs/framework-0063-notification-system.md
```diff
@@ -0,0 +1,422 @@
...
+
+#### Notification examples
+
+| Action             | Who triggers   | Who receives                                                                            |
+| ------------------ | -------------- | --------------------------------------------------------------------------------------- |
+| Comment create     | Creating user  | All involved in conversation                                                            |
+| Comment edit       | Editing user   | All involved in conversation                                                            |
+| Comment delete     | Deleting user  | All involved in conversation                                                            |
+|                    |                |                                                                                         |
+| Submission create  | Creating user  | Community owners/managers                                                               |
```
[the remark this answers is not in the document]
slint: Indeed the part about potentially triggering multiple notifications because of chained service calls under the same UoW has yet to be addressed... Thankfully we're using the UoW, so we can do some checks to de-duplicate.

The rule of thumb is to try to send notifications in the highest level service method possible. Now that we also have the auto-accepting community inclusion requests feature, we'll also have to see what notifications would get triggered and what is the desired behavior (e.g. maybe we want to notify other community curators that the record was auto-accepted/added).

rfcs/framework-0063-notification-system.md
```diff
@@ -10,231 +10,140 @@ Implementation details for a notification system to be used in InvenioRDM.
...
+## Detailed design
 
-    # Unique backend identifier ("email", "slack", "cern-notification")
-    id: str
+In this design, we describe a new `invenio-notifications` module based on the following:
 
-    # Options/parameters for the backend to deliver the notification
-    params: dict
-```
+- Simple "client" for triggering notifications from the service layer with low
```
TLGINO, line 45: Is a bit of the sentence missing here?

PR #84, opened 2024-09-09
vocabularies: add deprecation

jrcastro2:
DIscussion points:

- I don't think we need a confirmation modal here, WFYT?
- Should there be an option to use one of the superseeded values to replace the existing records if the user wishes to?
- How is the `deprecated_by` used ?
- Do we actually want to allow revert deprecation? if yes, should we call it revert or activate?
- When superseeding, should we allow to create a new record that would superseed it? (I understand this is a bigger task)
- Bulk deprecate?
- Should we display in other places that the values are deprecated (f.e. landing page)?
- Is the current model defined in the RFC good enough?
```json
{
...
    "system": {
        "deprecated_by": ["<id>", ... ],
        "deprecated_at": "2024-03-20T12:00:00Z"
    },
    "tags": ["deprecated", "depositable", "linkable"],
...
}
```


rfcs/rdm-0077-vocabulary-harvesting.md
```diff
@@ -246,6 +250,14 @@ Multi-step operation:
 
 #### Updating of relations
 
+#### Deprecation Behavior 
+
+When a property is marked as deprecated, it will be excluded from search results and cannot be selected for new records or modifications of existing records. The `deprecated_by` field is optional and can be used to indicate one or more values that supersede the deprecated property. Note that populating the `superseded_by` field does not affect existing records; those records will remain unchanged.
```
jrcastro2, line 255: There are some use cases where we might want to be able to select to deprecated values, for this, we allow to add deprecate values through the API for instance or and admin shuold be able to do it.

- [ ] Add a new tag "non-searchable" (or find a better name) that will prevent the values from appearing in the searches.
- [ ] Permission based, to allow admins to curate the records, only admins shuold see the deprecated values.
- [ ] just add the tag for the moment. no extra system fields.

PR #87, opened 2024-09-23
added collections RFC

alejandromumo:
closes https://github.com/zenodo/zenodo-rdm/issues/979


rfcs/rdm-0079-collections.md
```diff
@@ -0,0 +1,337 @@
...
+                                     
+In summary, below are the key points of this design along with the locations of the relevant discussions:
+
+- [🔑 Database considerations](#ctrees)
+- [🔑 Indices](#indices)
+- [🔑 Queries composition](#queries)
+                                     
+### Design
+    
+At a glance, a collection can be thought of as a set of records. To do so, we further expand the definition into:
```
lnielsen, line 64: A key point is that collections are **queries** over all records. A "set of records" I understand as an explicit assignment of a record to a collection which is what communities is.

rfcs/rdm-0079-collections.md
```diff
@@ -0,0 +1,337 @@
...
+
+Collections can be configured per community. I.e. for the European Repository it is interesting to have collections based on the records funded by the European Comission, while for a Biodiversity Repository it could be interesting to have these collections on different species (e.g. ants, insects, etc).
+
+For instance, see below multiple collections that exist under the community "European Open Repository":
+
+![](0079/1.png)
+    
+### Technical implementation
+
+Naturally, collections are organized in tree-like structures. I.e. Cern Document Server (CDS) defines multiple top-level collections [1] that each have, potentially, other collections attached.
```
lnielsen, line 85:
```suggestion
Naturally, collections are organized in tree-like structures. I.e. CERN Document Server (CDS) defines multiple top-level collections [1] that each have, potentially, other collections attached.
```

rfcs/rdm-0079-collections.md
kpsherva: maybe adding "transversal queries" would help?

rfcs/rdm-0079-collections.md
```diff
@@ -0,0 +1,337 @@
...
+        Super-Resolved SAR Data   <- Record
+```
+                                     
+For instance, collections can provide valuable insights of the contents of a community based on its records metadata, e.g. the funding metadata. This provides a more unique way of navigate through communities and analyse its records.
+
+## Motivation
+
+- As a community manager, I want to know how many records of my community are funded by the same grant
+- As a community manager, I want a high-level view that provides insights on the community records
+- As a researcher, I want to see all the records which are in a subject
```
kpsherva, line 50:
```suggestion
- As a researcher, I want to see all the records which are in a subject
- As instance manager I want to advertise records on a selected keyword
```

rfcs/rdm-0079-collections.md
```diff
@@ -0,0 +1,337 @@
...
+![](0079/4.png){width=1000}
+
+Another example of collections:
+    
+![](0079/5.png){width=500}
+*Euro Science subjects collections - http://publications.europa.eu/resource/dataset/euroscivoc*
+
+This solution proposes collections to be modelled as generic trees, whose information is persisted in a database. 
+    
+That said, collections data model will be based in the "Materialized Path" pattern. Read this [documentation from MongoDB](https://www.mongodb.com/docs/manual/tutorial/model-tree-structures-with-materialized-paths/) for a good overview of this pattern. The main concept is that each node (row in the DB) stores its `path` from the root to itself, excluding. For our example, it looks as follows:
```
kpsherva, line 101:
```suggestion
That said, collections data model will be based in the "Materialized Path" pattern. Read this [documentation from MongoDB](https://www.mongodb.com/docs/manual/tutorial/model-tree-structures-with-materialized-paths/) for a good overview of this pattern. The main concept is that each node (row in the DB) stores its `path` from the root to itself, excluding (missing part of the sentence here). For our example, it looks as follows:
```

rfcs/rdm-0079-collections.md
```diff
@@ -0,0 +1,337 @@
...
+
+**Database tables**
+
+We identified three new tables for this design:
+                                     
+1) `Collection`
+2) `CollectionTree`
+3) `CommunityCollectionTree`
+                                     
+![](0079/6.png)
```
kpsherva, line 121: from this diagram I understand we can have
- many collection trees ( I understand that it is for community purposes, right?)
- one collection can belong to only one tree ---> meaning we might have duplicated objects in different trees, are we OK with that?
- one community can have many collection trees -> what use case does it fulfill ? why not one to one?

rfcs/rdm-0079-collections.md
```diff
@@ -0,0 +1,337 @@
...
+The idea behind splitting `CollectionTree` from `Collection`, is to be able to define "root" collections that do not hold any query and simply group collections under a semantic umbrella. In our previous example, it would be modelled as such:
+                                     
+```
+European Open Repository  <- Community
+    Subjects                  <- CollectionTree
+        Natural sciences          <- Collection
+        Physical sciences         <- Collection
+```
+
+An important aspect of this design is that the `CommunityCollectionTree` allows the `community_id` to be set to `null`, indicating that the collection can be treated as a `root` collection. This enables the retrieval of all records from the repository without restricting them to a specific community.
```
kpsherva, line 139: I am a bit lost on the null collection id. Since the communityCollectionTree table seems to be intermediary table between the collection tree and community, then can we simply skip creating an entry in CommunityCollectionTree when we define a tree with no collection? why to allow to set it to null ? unless this is not "relational" model in a standard meaning

rfcs/rdm-0079-collections.md
```diff
@@ -0,0 +1,337 @@
...
+                                     
+<a name="ctrees">🔑</a> All the collections must be linked to a `CollectionTree` for organization purposes, however the display of the `CollectionTree` `title` on the UI could be optional. In a future iteration one could add a column to control the `CollectionTree` visibility.
+                                     
+<a name="indices">🔑</a> Apart from the usual PK, FK indices, the following indices are to be added to increase the performance of the most common used queries:
+                                     
+1) Index on `Collection.path`, possibly containing the `depth` column to allow matching only subcollections until a certain depth.
```
kpsherva, line 145: do we define any requirements about the allowed depth? I remember we mentioned this during the discussion

rfcs/rdm-0079-collections.md
```diff
@@ -0,0 +1,337 @@
...
+                                     
+1.  a record that belongs to a subcollection is guaranteed to also be included in its parent(s) collection(s).
+2.  resolving the query requires `N` primary key lookups on the `collections` table, where `N` is the node depth in the tree (e.g. 2 for `Particles` in the previous example)
+                           
+**Collection endpoints**
+
+We identified two new endpoints:
+
+1. `/<community_slug>/collections`
+2. `/<community_slug>/collections/<collection_slug>`
```
kpsherva, line 169: be on a lookout for creating URLs to add new collections
we ended up with `/commmunities/new` and `communities/search` and having to retrict "new" and "search" from possible slugs

rfcs/rdm-0079-collections.md
```diff
@@ -0,0 +1,337 @@
...
+
+The endpoint for a non-community collection is not yet defined, and will be listed as an open question in this document.
+
+### Configuration
+
+Initially collections entry point in the UI will be toggleable, e.g. disabled or enabled. However, this will not avoid the creation of the Database tables.
```
kpsherva, line 175: should this be a module then? if needed we install it and then it adds the tables?

rfcs/rdm-0079-collections.md
```diff
@@ -0,0 +1,337 @@
...
+    
+- for example, fetching all descendants of a node (collection) would require more complex logic than a simple text match query. 
+- although recursive queries can retrieve all the ancestors of a node with similar performance to a materialized path, they are more difficult to comprehend and implement.
+
+## Unresolved questions
+    
+1. **Non-community collections**: the endpoint is not yet defined.
+
+2. **Support alternative query strategies**: While we have defined query concatenation as the default approach, there may be other use cases where the query should be composed as a union of all descendant queries.​
+
```
kpsherva, line 332: this document is not touching the notion of collections built on facets, which we have discussed, is it on purpose?

rfcs/rdm-0079-collections.md
alejandromumo: These are good questions, I will do my best to reply:

1. The table `CommunityCollectionTree` does not make sense anymore since we just want `A community has N collection_tree` and `A collection_tree belong to ONE community only`. So the table `CommunityCollectionTree` disappears - I will update the RFC afterwards.

2.  a `null` `community_id` means that the collection belongs to the whole repository and not only a community. An entry on `CollectionTree` is still needed, to define the `root` node of that tree.

rfcs/rdm-0079-collections.md
alejandromumo: We had the depth in mind when modelling collections. This model (materialized paths) works with any depth, and we added the column to speed up queries for `depth > 1`. Any requirement / restrictions can then be added at the API or service level. 

As a side note, there's an optimized case, which is when `depth=1` (e.g. you want to query only the children of a node). The query is much faster since it does an exact text match:

```python
# Foo is the root node
foo = Collection("foo", parent=None)

# The children of 'foo' have the exact following path:
foo_children_path = ",foo,"
```
This can be used in DB like:

```sql
SELECT * collections
WHERE path = ",foo,"
```

rfcs/rdm-0079-collections.md
alejandromumo: It's a good point. 

Dicussed with Lars and the collection endpoints should be namespace by `tree`, otherwise there could be clashes with slugs. 

So we're moving to the following endpoint:

``` python
[
    "/<community_slug>/collections/<tree_slug>/<collection_slug>"
]
# examples:
# /eu/collections/funding_programmes/horizon2020
# /eu/collections/subjects/mathematics
```

rfcs/rdm-0079-collections.md
alejandromumo: Facets per collection needs more work and iterations, because storing facets in the DB can lead to weird errors if e.g. a facet is updated. There are other questions, e.g. `where does the collection facet come from? ` - it could be from itself or from parents in the tree.

You are right, it should be mentioned in the RFC and added to the `Open questions`.

rfcs/rdm-0079-collections.md
alejandromumo: Personally, that was my first thought. But since collections are tightly linked to records (and communities), we will add them in `invenio-rdm-records`. So to answer your question, one can't opt-out from having the tables

rfcs/rdm-0079-collections.md
kpsherva: thanks for the answers, I have follow-up questions:

2. by duplicated objects I meant we might have a collection with query "subjects:bla"  under community EU and "subjects:bla" collection under community BLR (each collection with different id, but the query is the same). My question was if we are aware of this and accepting it.
3. have you considered the community query to be the root collection and the grouping is simply lvl 2 of the tree?

ntarocco: @alejandromumo we discussed point 2, we can close it.
However, n. 3 is a valid point, which was also already discussed in the past. What if a community does not need a `tree`, but would simply like to create a set of collections?
We could have a "default" tree, and maybe hide from the UI, however, in the URL, the tree slug is required. I think that this deserves a quick chat.

ntarocco (commented):
The main question to answer is if and how we want to allow having collections without a collection tree. If not, it won't be easy to explain this to the community curators.
I have the feeling that, most of the time, community curators will create only one collection tree.

rfcs/rdm-0079-collections.md
alejandromumo: Currently, a `tree` is mandatory to group the collections. But indeed we could have a `default` tree per community, so users don't have to do it every time they need a collection.

It could be e.g. `/communities/<slug>/collections/default/<collection>` where `default` is the `tree.slug` 

WDYT? ping @lnielsen

lnielsen: Let's discuss IRL tomorrow. If I understand correctly, the main issues is the URL pattern if you only have one collection tree in a community? I.e. currently you need ``/communities/<id>/collections/<tree-id>/<collection-slug>`` vs you would like to remove ``tree-id`` so it becomes ``/communities/<id>/collections/<collection-slug>``

My five cents is I agree it would be nice to avoid the tree id when only one tree exists, and possibly we could simply allow the tree slug to be empty (or mark the tree as the default one). I makes the resolution of a URL a little bit more complex and there's some consistency checks we would need to take care of (e.g. ensuring that a tree id and community slug is not overlapping)

rfcs/rdm-0079-collections.md
```diff
@@ -0,0 +1,349 @@
...
+![](0079/4.png){width=1000}
+
+Another example of collections:
+    
+![](0079/5.png){width=500}
+*Euro Science subjects collections - http://publications.europa.eu/resource/dataset/euroscivoc*
+
+This solution proposes collections to be modelled as generic trees, whose information is persisted in a database. 
+    
+That said, collections data model will be based in the "Materialized Path" pattern. Read this [documentation from MongoDB](https://www.mongodb.com/docs/manual/tutorial/model-tree-structures-with-materialized-paths/) for a good overview of this pattern. The main concept is that each node (row in the DB) stores its `path` from the root to its own node, excluding itself. For our example, it looks as follows:
```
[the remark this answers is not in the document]
alejandromumo: you're right, I had two examples but one was removed. I updated the RFC accordingly

PR #99, opened 2025-07-01
refactor: move rdm rfcs into rfcs folder

carlinmack:
:heart: Thank you for your contribution!

there are RDM RFCs in this folder, so as far as I can tell it's just historical that no one has consolidated them yet


fenekku (approved):
Yeah, I think we were splitting hair a bit too much at the start. All in one place , with the prefix of the most relevant scope for the RFC sounds like the right ... *move* :wink:

PR #109, opened 2025-10-27
framework+rdm: establish Python support policy

fenekku:
This is actually my summary of the discussions maintainers had. I've separated out my own comments / thoughts, so this should be understood.

Also in the below text "supported Python version" should be read as "officially supported by the Python foundation/community"


rfcs/rdm-0109-python-versions.md
```diff
@@ -0,0 +1,137 @@
...
+
+## Summary
+
+This RFC outlines the Python support policy for InvenioRDM and the related Invenio ecocystem. In particular, it outlines how it relates to concrete implementations/use-cases. Some aspects are left flexible for agile decision making in future contexts.
+
+The general idea is:
+
+|                                 | Minimum                                       | Maximum                             |
+| ------------------------------- | --------------------------------------------- | ----------------------------------- |
+| Invenio framework level modules | Keep current minimum                          | Latest version officially supported |
```
fenekku: From my notes I am not sure if we meant keep those to the smallest Python version supported or to the current version they support. Some of them support 3.8 and other long deprecated versions...  My thoughts on that "Keep current minimum" are there: https://github.com/inveniosoftware/rfcs/pull/109/files#diff-ac3af67c596e1c277c361a8934a259cddd2c7be3fb9c1e6001a2f3ac5326d352R93-R100 .

rfcs/rdm-0109-python-versions.md
```diff
@@ -0,0 +1,137 @@
...
+Invenio-wide modules like invenio-app,-search... (as opposed to InvenioRDM-specific ones like invenio-records-resources,-rdm-records ...) are not driven by the InvenioRDM cadence and should be stable. They should keep supporting the minimum Python version they did and be independent of RDM.
+
+[Mine]
+It seems strange to keep them supporting versions of Python that are deprecated however. I could see us moving that minimum in tandem with officially supported version or with InvenioRDM cadence actually. Would simplify development concerns and encourage users to switch to newer Python versions to keep receiving patches/new features.
+
+It also means that those modules are the ones that will "cost" us most in terms of maintenance:
+- they cannot benefit from greater mimimum as RDM modules
+- they require adaptive code to support their now deprecated parts
+- If minimum changed every year that is very frequent
+[end mine]
```
fenekku: My own thoughts there.

rfcs/rdm-0109-python-versions.md
```diff
@@ -0,0 +1,168 @@
...
+- ... take advantage of Python performance gains as a baseline
+- ... don't have to change minimum version every year and constantly churn
+- ... are alerted and account for upcoming deprecations in development as soon as official next latest Python version is out
+- ... align Python and InvenioRDM version numbers which is nice
+
+### InvenioRDM — periods of overlap with deprecated Python versions
+
+The Python support policy outlined above leads to a scenario where a newly released InvenioRDM and the previous, but still supported for 6 months per [RDM support policy](https://inveniordm.docs.cern.ch/releases/maintenance-policy/), do not share the same mimimum required Python version for that overlap period. This could be a rare occurrence (e.g., every 5 years), but it has to be accounted for.
+
+As is, it would mean that during that period of time, any fix would have to be compatible with the previous minimum Python version — a different and unsupported (by Python community) one.
```
fenekku: This is taking the "first option" mentioned in previous version of this RFC: 

Maintainers continue for 6 months to make fixes and backports that are compatible with the previous minimum Python version. This means being more careful about fixes and it doesn't promote adoption of a secure/supported version of Python. It is simpler because it doesn't require anything else. Past the 6 months mark, all the new features of the new minimum Python version can be adopted (and all deprecation workarounds that can be dropped are).

slint (commented):
Thanks for putting this down in writing, I really like how this is forming :)

Had a go at the "core" part of the RFC regarding supported versions across `Invenio{,RDM,ILS}`. I didn't have time/stamina to go through the transition periods, but I'll have a go again tomorrow.

fenekku:
I pushed some edits. 

Areas of discussion left (I can be swayed and it would be nice to hear from others too):

- Concretize the Officially Supported versus Compatible version concept: are they different enough in practice in other ways? what do maintainers think about them?
- Only 1 officially supported version at a time or could there be multiple? Effort to maintain?   

I will recheck at the end of my day else when I come back from holidays (I am fine if a decision about this is done while I am away. Either way it shakes out, there is nothing I strongly disagree with).

rfcs/rdm-0109-python-versions.md
```diff
@@ -0,0 +1,198 @@
...
+| Python 3.15 release<br> 3.10 deprecation | invenio-app-rdm | [3.14, 3.15]<br> (3.15 is an option, and could have not been there) | InvenioRDM-specific module<br> active anointed version is 3.14<br> highest official version is 3.15 |
+| Python 3.15 release<br> 3.10 deprecation | invenio-app | [3.11, 3.14, 3.15]<br> (3.15 an option, not a must) | InvenioRDM-ecosystem module<br> active anointed version is 3.14<br> lowest and highest official versions are now 3.11 and 3.15 |
+| InvenioRDM v15 Release<br> anointed version kept the same | invenio-app-rdm | [3.14, 3.15] | InvenioRDM-specific module<br> active anointed version is 3.14<br> highest official version is 3.15 |
+| Python 3.16 release<br> 3.11 deprecation | invenio-app-rdm | [3.14, 3.16] | InvenioRDM-specific module<br> active anointed version is 3.14<br> highest official version is now 3.16 |
+
+Cases where different anointed versions overlap are covered in the "Transitions" section.
+
+### Meta markers (requires-python / python_requires and co.)
+
+`requires-python` (`pyproject.toml`) or `python_requires` (`setup.cfg`) *should not* be present in any module anymore.
```
fenekku: This is the only part I am personally more ambivalent about, because it makes the cutoff version implicit. 

- ✔️ 1 less thing to keep up-to-date. 
- ❌ uv/pkg managers can't use that info to provide appropriate python-version-supporting package version to client

Alternative there would be to put the lowest official version by default and replace it by lowest anointed version if any anointed version specific features are used (I don't think we should be any more granular than this).

Samk13: I see the maintenance argument, but I share some of your concerns here.

One thing `requires-python` gives us today is a machine-readable compatibility contract that packaging tools can use during dependency resolution. 
Without it, users on unsupported versions may end up installing a package successfully and only discover incompatibilities later when running tests or the application itself.

I agree that maintaining exact compatibility markers can become a bit burdensome, especially if the policy is to support a moving set of "anointed" versions. However, I wonder if there is a middle ground where we keep a conservative lower bound (e.g. the oldest officially supported version) while avoiding upper bounds and frequent updates. That would still provide useful metadata to tooling without requiring us to encode the full support matrix in package metadata.

Could you elaborate a bit more on why the loss of packaging-tool support is considered an acceptable trade-off here?

fenekku: My understanding of @slint arguments on this was: `require-python` (and equivalent) would add maintenance cost and would potentially unnecessarily cutoff lower Python versions than the anointed one (and I *think* I even heard something akin to: it would cutoff package users not even using an official Python version anymore). To flesh out the cutoff argument: by not having it, packages that are still compatible with lower Python versions than the anointed would still be installable, and *only* incompatible ones would have to be forked. It's a "as permissive as possible" approach.

I think for me the higher priority goal would be to focus on delivering a piece of software built on top of/relying on supported software as much as possible. Obviously not all our package dependencies are maintained, but big ticket items like Python, the database, the web servers, docker are important. The legacy-compatibility objective would in effect be achieved to a lesser degree this way: folks who don't want to upgrade anything can keep using the old stuff and it should keep running with all the compatible old stuff. Problems mostly arise when folks do partial upgrades, and then I would argue they need to partially downgrade or move to the new version wholesale. 

Even just having the require-python be `>= <lowest official Python version>` would be fine by me. Using the anointed version when we know nothing below is compatible would be even nicer.

